PR #3, opened 2023-10-30
feat(github): initial github templates

jeffchew:
This sets up various github tempaltes for the repo.


mattrosno:
@jeffchew it doesn't hurt to also have the files in this repo, but code of conduct and support didn't have to be set here, as by default this repo would inherit the org's code of conduct and support statements from here:

https://github.com/carbon-design-system/.github/tree/main/.github

PR #45, opened 2023-12-07
feat(feedback): added feedback component

priyanshu891:
Closes #

added feedback container to help user to record or give feedback on generated content.

#### Changelog

**New**

- added feedback container 

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

N/A


priyanshu891:

priyanshu891:
recheck

packages/chat/components/feedback/src/feedback-container.template.ts
```diff
@@ -0,0 +1,323 @@
...
+ * Feedback Container component to record and give feedback on AI generated content
+ */
+export class FeedbackContainer extends LitElement {
+  static styles = styles;
+
+  /**
+   * API Key of registered application for using this component
+   */
+  @property({ attribute: 'api-key' })
+  private _api_key: string = '';
```
annawen1, line 35: You can define the type in the `@property({ type: String, ...})` instead. Is it necessary to make the

packages/chat/components/feedback/src/feedback-container.template.ts
```diff
@@ -0,0 +1,323 @@
...
+  /**
+   * Method for toggling the Feedback Modal
+   */
+  _toggle() {
+    this.isModelOpen = !this.isModelOpen;
+  }
+
+  handleSlotChange() {}
+
+  render() {
```
annawen1, line 316: We want to break the component out to separate parts so that adopters who want to extend the component can take in certain parts and adjust as needed.

If you take a look at the `test-input` component, we have `packages/chat/components/test-input/src/test-input.ts` file that contains all the class properties and method. We then import the class into the `packages/chat/components/test-input/src/test-input.template.ts` file which contains what you would put in the `render()`. And finally we have the `packages/chat/components/test-input/test-input.ts` file that take both the class file (`src/test-input.ts`) and template file (`src/test-input.template.ts`) and registers the component with the `@customElement()` decorator.

So for this `feedback-container`, let's break this `feedback-container.template.ts` file out 

`src/feedback-container.ts` -> class methods and properties
`src/feedback-container.template.ts` -> the render()
`feedback-container.ts` -> bring in both files - view `test-input.ts` for reference

annawen1:
Please add the import to the `feedback-container` to the `packages/chat/index.ts` file

annawen1 (approved):

PR #59, opened 2024-01-12
First commit of chat component

ocornec:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


packages/chat/react/feedback.ts
```diff
@@ -1,19 +0,0 @@
-/**
```
annawen1, line 1: Can you put both the react/feedback.ts and react/test-input.ts files back in?

packages/chat/package.json
```diff
@@ -35,7 +35,7 @@
   "dependencies": {
     "@babel/runtime": "^7.23.2",
     "@carbon/ai-utilities": "0.0.1-rc.1",
-    "@carbon/web-components": "2.1.1",
```
annawen1, line 38: Can you bump this back up to 2.1.1?

packages/chat/components/chat/src/chat.scss
```diff
@@ -0,0 +1,272 @@
...
+@use '../../../../../globals/scss/vars' as *;
+@use '@carbon/styles/scss/theme' as *;
+:host(#{$c4ai-prefix}-chat) {
+  position: relative;
+  display: block;
+  block-size: 320px;
+  font-family: 'IBM Plex Sans', sans-serif;
+  min-block-size: 320px;
+
+  .chat-container {
```
annawen1, line 17: In Carbon we follow the [CSS BEM (Block Element Modifier) naming convention](https://getbem.com/introduction/) and use prefixes in the classnames: 

Example here of the text-input component: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/styles/scss/themes/expressive/components/_text-input.scss#L26

packages/chat/components/chat/src/chat.template.ts
```diff
@@ -0,0 +1,99 @@
...
+    _messageText: messageText,
+    _queryInProgress: queryInProgress,
+    _handleHoverIn: handleHoverIn,
+    _handleHoverOut: handleHoverOut,
+    _handleEdit: handleEdit,
+    _handleFeedback: handleFeedback,
+    _handleAPIselection: handleAPIselection 
+  } = customElementClass;
+
+    return html`<div class="chat-container">
```
annawen1, line 39: You'll need to import the following:
```
import { settings } from '@carbon/ai-utilities/es/settings/index.js';
const { stablePrefix: c4aiPrefix } = settings;
```
and you'll be able to use the prefix in the classnames ie:
```
<div class="${c4aiPrefix}--chat__container" ...
```

packages/chat/components/chat/src/chat.ts
```diff
@@ -0,0 +1,355 @@
...
+ */
+export default class C4AIChat extends LitElement {
+  static styles = styles;
+
+   /**
+   * Core array of message objects to display and interact with, invidual messages are currently described as:
+   * {"message":"displayed text", "origin":"can be bot or user, depending on user-defined role names"}
+   */
+  private _messages: any[] = [];
+  /**
```
annawen1, line 29: Nitpicking here but can you place a new line between all the property/state declarations here so it's easier to read?

packages/chat/components/chat/src/chat.ts
```diff
@@ -0,0 +1,355 @@
...
+   **/
+  _checkForCode(string) {
+    console.log(string);
+    console.log(string.includes('```'));
+    if (!string.includes('```')) {
+      return [{ type: 'text', text: string }];
+    }
+    const strings = string.split('\n');
+    let mode = 'text';
+    // commenting below because I don't see where `previousMode` is being used in the code
```
annawen1, line 213: I had added these comments and commented out  `previousMode` variable when resolving the TS errors you ran into in the beginning - can you double check to see if they are needed and are used somewhere?

packages/chat/services/LlamaPlugin/LlamaPlugin.ts
```diff
@@ -0,0 +1,203 @@
...
+        "projectId": project_id,
+        "promptData":{
+            "imput": history,
+            "modelId": model_id,
+            "modelParameters": parameters
+        }
+    }
+
+    console.log(watsonxPayload)
+    console.log(_watsonxai_key)
```
annawen1, line 69: do we want to be console logging this info?

packages/chat/components/chat/__stories__/chat.stories.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const Default = {
+  render: () => html` 
+    <c4ai-chat
+      model="llama-2"
+      userprompt="You are Watson, you will answer all my questions."
+      api-url="https://bam-api.res.ibm.com/v1/generate"
+      
+      temperature="0.0"
```
annawen1, line 26: Maybe we can utilize the storybook controls here for some of the properties? That way user can change/play with certain aspects of the chat right from storybook.

example with the extended-button: https://github.com/carbon-design-system/carbon-for-ai/blob/main/packages/extended-button/__stories__/extended-button.stories.js#L37

Theres different types you can have in the controls (Boolean, select - which you provide an array so it gives user dropdown options to choose from, label - text the user can input/edit, etc.)

annawen1:
ci-checks are failing because the lock file `pnpm-lock.yaml` is not up to date. Can you run `pnpm install` in the root and push up the updated lock file?

packages/chat/components/chat/src/chat.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    /*
+    let text = parent.querySelector(".message-text").innerHTML;
+    console.log(text)
+
+    const textArea = document.createElement('textarea');
+    textArea.className = 'editableTextArea';
+    textArea.value = text;
+
+    textDiv.replaceWith(textArea)*/
```
annawen1, line 335: Do we need this commented out section?

packages/chat/react/feedback.ts
ocornec: I pulled in with git fetch upstrean but don't see them, the version I pushed before was the latest so unsure where these files are, are they auto generated?

packages/chat/package.json
ocornec: done

packages/chat/components/chat/src/chat.scss
ocornec: done, but should the prefix be added to every single class I have in my css or just the container?

packages/chat/components/chat/src/chat.ts
ocornec: done, will keep it in mind in the future

packages/chat/components/chat/src/chat.ts
ocornec: it is! it dictates whether the current parsing mode is "string" aka plain text or "code", aka a code block. since the response contains both I need a value to toggle between both given a predefined delimiter. I really don't know why TS throws errors on this, lmk how to resolve this.

packages/chat/services/LlamaPlugin/LlamaPlugin.ts
ocornec: removed, used for testing the CORS issues I mentioned before, can only know once I'm testing the component on the official carbon labs ibm URL.

packages/chat/components/chat/__stories__/chat.stories.js
ocornec: Great idea, I'll implement that later.

packages/chat/components/chat/src/chat.ts
ocornec: it was an early version of text editing but unneeded since Priyanshu put in the text feedback component, removed it

packages/chat/components/chat/src/chat.template.ts
```diff
@@ -0,0 +1,99 @@
...
+        <div class="chat-messages">
+        &nbsp;
+          ${messages.map(
+            (message,index) => html` 
+              <div class="${message.origin}-message subtype-${message.style}" @mouseout="${(e) => handleHoverOut(index,e,message.origin)}">
```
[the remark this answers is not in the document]
ocornec: That's something I was struggling with but given the complexity of messages (rendering text, code, images, urls, UI actions, arrays of each..) I do see the need to extrapolate it into a subcomponent. my next push will take care of it.

packages/chat/components/chat/src/chat.scss
annawen1: this should be done for every single class - you can use the carbon-for-ibm-dotcom repo as a reference: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/packages/carbon-web-components/src/components/text-input/text-input.scss

packages/chat/components/chat/src/chat.ts
annawen1: TS is throwing the error because you created the variable and set a value, but it's not detecting where in the code you are utilizing the variable. Where in the code are you checking the `previousMode` value?

annawen1: I see you set it with the initial "text" value and then set it to the `mode` value, but is `previousMode` passed into an object somewhere? I'm not seeing where it's used

packages/chat/services/LlamaPlugin/LlamaPlugin.ts
annawen1: The carbon labs ibm url - the storybook link for that is `https://ai-canary.carbondesignsystem.com/?path=/docs/components-chat-test-input--docs` so not under the ibm.com domain...not sure if that will work?

packages/chat/react/feedback.ts
ariellalgilmore: I belive anna put them in manually. You should be able to just copy and paste the contents from github where it shows they were removed and push a commit with the files again

ocornec:

ariellalgilmore:
recheck

PR #95, opened 2024-03-04
Feat/chat storybook updates

ocornec:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- Added Card object for urls and videos, resizable window, added list extraction logic and subelement

**Removed**

- {{removed thing}}

#### Testing / Reviewing


packages/chat/components/header/__stories__/header.stories.js
```diff
@@ -0,0 +1,26 @@
...
+  tags: ['autodocs'],
+};
+
+export const Default = {
+  /**
+   * Renders the template for Storybook
+   *
+   * @returns {TemplateResult<1>}
+   */
+  render: () => html` <c4ai--chat-header> </c4ai--chat-header>`,
```
annawen1, line 25: just noticed we should not being using 2 hyphens, let's keep it consistent with one

```suggestion
  render: () => html` <c4ai-chat-header> </c4ai-chat-header>`,
```

annawen1 (approved):
Needs changes for sure, but we can revisit afterwards

PR #179, opened 2024-06-05
feat(tag): New Tag component

lily-peng:
Closes [#219](https://github.ibm.com/semantic-automation/watsonxdata-techpreview/issues/219)

A tag component that is used to represent a data table.

Examples:
image

image

Features:
- Customize inner text
- Customize left border color
- Displays Carbon Tooltip on hover
- Customize tooltip position & text
- Fires click event

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing


packages/tag/components/tag/src/tag.ts
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Left border color
+   */
+  @property({ attribute: 'color', type: String })
+  color;
+
+  /**
+   * Tooltip position
+   */
+  @property({ attribute: 'tooltipPosition', type: String })
```
annawen1, line 36: with attribute names we usually separate by hyphen 

ie. `tooltip-position`

packages/tag/components/tag/src/tag.template.ts
```diff
@@ -0,0 +1,45 @@
...
+    tooltipPosition: tooltipPosition,
+    tooltipText: tooltipText,
+    handleClick: handleClick,
+  } = customElementClass;
+
+  return html` <div class="${clabsPrefix}--tag">
+    <div class="${clabsPrefix}--tag-container">
+      <cds-tooltip align="${tooltipPosition}">
+        <button
+          class="tag sb-tooltip-trigger"
```
annawen1, line 35: For classnames, we'll add a prefix as well like you've done for the other classes

`${clabsPrefix}--tag`

The `sb` in `sb-tooltip-trigger` usually stands for storybook so unless this is only used in storybook, I would recommend making the classname more specific to the component itself (ie. `${clabsPrefix}--tag-tooltip-trigger`

packages/tag/__stories__/tag.stories.js
```diff
@@ -0,0 +1,76 @@
...
+  argTypes: defaultControls,
+  args: defaultArgs,
+  /**
+   * Renders the template for Storybook
+   * @param {string} args.content - content to generate from
+   * @returns {TemplateResult<1>}
+   */
+  render: ({ text, color, tooltipPosition, tooltipText, onClick }) =>
+    html` <clabs-tag
+      .color=${color}
```
annawen1, line 70: if you're passing in a string to the attribute, you can pass it in normally 

`color=${color}`, the `.` is only used when you are passing in an object to the attribute of a lit component

.gitignore
```diff
@@ -28,4 +28,5 @@ custom-elements.json
 *.sln
 *.sw?
 .nx/*
-agentic.stories.js
\ No newline at end of file
+agentic.stories.js
+/.history
```
annawen1, line 32: where is this `history` folder coming from?

packages/tag/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "build": "gulp build --option tag",
+    "build:dist": "rm -rf dist && rollup --config ../../tools/rollup.config.dist.js",
+    "build:dist:canary": "rm -rf dist && rollup --config ../../tools/rollup.config.dist.js --configCanary"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.23.2",
+    "@carbon-labs/utilities": "0.8.0",
+    "@carbon/grid": "^11.21.0",
+    "@carbon/web-components": "2.7.0",
+    "uuid": "^9.0.1"
```
annawen1, line 40: We can probably remove the `uuid` package since I don't see it used in this tag package
```suggestion
```

.gitignore
lily-peng: they're from my VSCode settings; I'll remove the line from the .gitignore

packages/tag/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@carbon-labs/ai-tag",
+  "version": "0.8.0",
```
annawen1, line 3:
```suggestion
  "version": "0.0.1",
```

packages/tag/components/tag/src/tag.template.ts
```diff
@@ -0,0 +1,45 @@
...
+    handleClick: handleClick,
+  } = customElementClass;
+
+  return html` <div class="${clabsPrefix}--tag">
+    <div class="${clabsPrefix}--tag-container">
+      <cds-tooltip align="${tooltipPosition}">
+        <button
+          class="${clabsPrefix}--tag-button ${clabsPrefix}--tag-tooltip-trigger"
+          .color=${color}
+          .tooltipPosition=${tooltipPosition}
```
annawen1, line 37: oh just noticed this here too!

```
color=${color}
tooltip-position=${tooltipPosition}
```
```suggestion
          color=${color}
          tooltip-position=${tooltipPosition}
```

mjabbink:
— The colors need to be limited to be accessible. 
— It needs light and dark theming. We would not want to encourage ALL CAPITAL LETTERING unless this is user-generated content, and we have no control or influence over that content. 
— Size (height) and spacing specs. Font size and weight, etc.
— What do the colors mean? 
— Would there be any other actions? Can you delete them?

lily-peng:
> The colors need to be limited to be accessible. What do the colors mean?

Hi @mjabbink, we kept the left side color open for the user to input whatever they want for flexibility. The current use case for the Watsonx Proof of Technology is for the Tag to represent a data table, with summary information in the tooltip on hover & when clicked. The left side color would be a way to differentiate different types of tables. I'm not sure about what the different types could be though, @daniel-karl would be better to answer these questions.

> ALL CAPITAL LETTERING

Sorry, I'm not sure what this is referring to?

> Would there be any other actions? Can you delete them?

No deleting as far as I know, and only actions are hover and click.

annawen1 (approved):

mjabbink:
@lily-peng The colors alone are not accessible. e.g. a color blind person might see the difference between the colors, so they wold need another way to distinguish between the “types” of tags. If the name of the tag does that then ok, but if the name does not make it clear, not will the color help. The user wold also need to know what the color meant and if changing this cold also pose a problem.

PR #332, opened 2024-12-13
chore(chat): Code functionailty fixes, Chart updates, Formula render fix

ocornec:
# @chore(chat): Code functionailty fixes, Chart updates, Formula render fix

## Decription

For Code:
- added line counts and language display
- enabled custom language setting
- extensive SCSS color theming expansion to handle COBOL FORTRAN C# Docker Lisp etc...
- docs: tons of examples and demos + new options available
- mixing editing/ticks/titling to work together and resize properly
- added char/line thresholds to automatically disable coloring when limit is exceeded for performance

For Charts:
- New examples in docs
- Highlight mechanism
- Multi-chart fixes for sizing, legend scaling

For Footer: external re-focus control

For Formula: Targeting shadowDOM instead of global window to avoid DOM leakage

For Popup: Styling updates up to Design standards


packages/web-components/src/components/chat/components/formulaElement/src/formulaElement.ts
```diff
@@ -33,6 +33,36 @@ export default class formulaElement extends LitElement {
   @state()
   formula;
 
+  /**
+   * startMathJax -  edit target document for rendering
+   */
+  async startMathJax() {
```
ethanwinters, line 39: `import 'mathjax/es5/tex-mml-chtml.js';` still pollute the global "window" object. This means that if the chat is running along side other code that uses mathjax (not a crazy thought) that they will clobber each other. Going to approve the PR, but we gotta work this out or move to katex.

ethanwinters (approved):

PR #358, opened 2025-01-07
feat(storybook): add root level storybook instance for landing page

kennylam:
This adds a root-level Storybook which pulls both Web Components and React Storybooks into the same instance. This is purely for Storybook deployment to stage/prod environments and not intended for local development.

image

#### Changelog

**New**

- add Storybook to project root

**Changed**

- add Storybook publish workflow for gh-pages


.github/workflows/storybook-publish.yml
```diff
@@ -36,6 +40,14 @@ jobs:
         working-directory: 'packages/react'
         run: yarn storybook:build
```
annawen1, line 42:
```suggestion
      - name: Create CNAME
        run: |
          touch storybook-static/.nojekyll
          touch storybook-static/CNAME
          echo "labs.carbondesignsystem.com" > storybook-static/CNAME
```

I think we might need this step if the root is gonna be cleared out and replaced with the root build files

annawen1 (approved):
LGTM!

PR #410, opened 2025-02-04
feat(SideNav): implement 3rd level navigation

### LINKED ISSUE #415: [Platforms] SideNav - support nested SideNavMenu inside SideNav panel
Design issue https://github.com/carbon-design-system/carbon/issues/18291

Currently the SideNav panel only supports SideNavLink children, we need to also support SideNavMenu when in a collapsed state.

```jsx
 <SideNav
  navType={SIDE_NAV_TYPE.PANEL}
  isChildOfHeader={false}
  hideOverlay
  aria-label="Product navigation">
  <SideNavItems>
    <SideNavMenu renderIcon={Fade} title="Product 4">
      <SideNavMenuItem href="#">Link 1</SideNavMenuItem>
      <SideNavMenuItem href="#">Link 2</SideNavMenuItem>
      <SideNavMenuItem href="#">Link 3</SideNavMenuItem>
    </SideNavMenu>
  </SideNavItems>
</SideNav>
```

### When SideNav panel is collapsed
- [x] Icons for SideNavMenu should link to first child SideNavMenuItem
- [x] Should not ben able to open/close menu's from collapsed state

### When SideNav panel is expanded
- [x] SideNavMenu should function as it currently does, open/close

### Examples
- [x] Update default and side nav panel stories
- [x] Update overview.mdx
- [x] Update example for Stackblitz

IgnacioBecerra:
Closes https://github.com/carbon-design-system/carbon/issues/18466
Closes #415 

This PR implements the 3rd level navigation for the SideNav component, as well as the Panel navType feature where the submenu icon will link to its first link when the sidenav is in a collapsed state. 

The treeview state is enabled when the prop `isTreeview` is added to the `SideNav`. If there is a nested sub-menu, this feature will be enabled automatically.

#### Changelog

**New**

- TreeView navigation for UIShell
- added SideNavMenu nesting feature
- added keyboard arrow navigation
- fixed styles for Carbon Labs UIShell
- sub-menu links to first link in a collapsed SideNav

#### Testing / Reviewing
Ran `npm test`, also test using keyboard navigation


alisonjoseph:
One thing I wanted to make sure was clear, we only want to keyboard nav to switch to treeview if it has a 3rd level, otherwise it should function as it does today with tabs.

alisonjoseph:
Copying over @oliviaflory comments  from https://github.com/carbon-design-system/carbon-labs/issues/419 related to the 3rd level nav here


### Styles


2. Sub menu level 1: Both icons should change to $icon-primary on hover

Image

3. Sub menu level 2: chevron should change to $icon-primary on hover

Image

4. Icon of parent menu should be $icon-primary when sub level link is selected

Image


## Side nav

### Keyboard

1. Cannot access the expand button via keyboard when the side nav is collapsed

2. Cannot access the collapse button via keyboard when the side nav is expanded

### Styles

1. Icon in expanded state should change to $icon-primary on hover

Image

2. Icon in collapsed state should change to $icon-primary on hover

Image

IgnacioBecerra:
PR is ready for review! All the new features we published are included, with fixes and improvements. Also fixed some style issues.

cc: @oliviaflory - please review when you have time next week! I addressed your comments from that bigger issue.

oliviaflory:
@IgnacioBecerra the styles are looking good, thanks for fixing those!

This is a bit nit-picky but anything that has additional items should probably be called `sub-menu`, we should reserve `link` for a single click target (no nested items). Seeing this in the [Default With Tree View story](https://deploy-preview-410--carbon-labs-react.netlify.app/?path=/story/components-uishell--default-with-tree-view) and some of the side nav stories, but maybe side nav isn't part of this PR
Screenshot 2025-02-10 at 10 40 31 AM

alisonjoseph:
Couple functional things I'm seeeing. I can't tab into SubMenuItems anymore in the SideNav without tree

https://github.com/user-attachments/assets/77e4efac-2180-4274-94ae-8761fe13a555


I'm also seeing an extra keyboard tab stop in the DefaultWithTreeView story. After hitting enter on the hamburger to open the menu, I have to tab twice to get to the first link, this doesn't happen with the non-tree sidenav.

alisonjoseph (changes requested):
Tabbing is working and the double tab issue is also gone now. I ran the a11y checker and there are several errors to look through. 

Screenshot 2025-02-11 at 7 55 57 AM



- I believe you will want to move role=tree to the ul and off of nav
- add an aria-label to the ul
- and for the tabbable error, I think its complaining about the buttons for the submenu all having tabindex="-1"

IgnacioBecerra:
@alisonjoseph Hm, adding another tabindex to the `<li role="treeitem">` would make it so the focus gets "doubled", since the real focus should be within the `<a>`. For the Treeview component, all the focus styles are in the `<li>`, so that's where it differs. Thoughts about this?

alisonjoseph:
@IgnacioBecerra I'm now wondering if the role=treeitem should actually be on the a tag/link instead of the li? That appears to be how they are doing it in the example you can open in codepen here https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/

alisonjoseph:
Couple other things, I'd like to add at least some minimal docs to the overview page for this before merging. 

- Update default story to use type treeview, can then remove the default with tree story
- Update stackblitz example to use type treeview
- Update overview page
  - Update main code sample
  - Add new section for treeview with a note around when to use type treeview (when more than 2 levels) with a code sample(can probably just use the code from sidenav with third level story).
  - Should we re-name sidenav with third level to SideNav TreeView?

alisonjoseph:
@IgnacioBecerra I'm fine with keeping the story name you got from Olivia 😄

emyarod:
> @IgnacioBecerra I'm now wondering if the role=treeitem should actually be on the a tag/link instead of the li? That appears to be how they are doing it in the example you can open in codepen here https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/
> 
> @emyarod thoughts on this. The navigation as a tree is a little different than the general treeview component because it has links.

yeah I believe the treeitem role goes on the link like in the w3 example

packages/react/src/components/UIShell/components/SideNavMenuItem.tsx
```diff
@@ -87,16 +89,16 @@ export const SideNavMenuItem = React.forwardRef<
       linkRef.current.style.paddingLeft = `${calcLinkOffset()}rem`;
     }
   }, []);
+  console.log(navType);
```
alisonjoseph, line 92:
```suggestion
```

packages/react/src/components/UIShell/components/styles/_side-nav.scss
```diff
@@ -48,12 +48,22 @@ div:has(.#{$prefix}--header)
...
-//----------------------------------------------------------------------------
+.#{$prefix}--side-nav__link,
+.#{$prefix}--side-nav__submenu {
+  &:hover {
+    .#{$prefix}--side-nav__icon > svg,
+    .#{$prefix}--side-nav__submenu-chevron > svg {
+      fill: $icon-primary;
+    }
+  }
+}
```
alisonjoseph, line 65:
```suggestion
.#{$prefix}--side-nav__submenu:hover {
  .#{$prefix}--side-nav__icon > svg,
  .#{$prefix}--side-nav__submenu-chevron > svg {
    fill: $icon-primary;
  }
}
```

super nitpicky, but we try to avoid nesting more than once in general in scss files in Carbon.

packages/react/src/components/UIShell/components/styles/_side-nav.scss
```diff
@@ -63,6 +73,14 @@ div:has(.#{$prefix}--header)
     margin-inline-end: $spacing-05;
   }
 
+  .#{$prefix}--side-nav__item.#{$prefix}--side-nav__link {
+    &:hover {
+      .#{$prefix}--side-nav__icon > svg {
+        fill: $icon-primary;
+      }
+    }
+  }
```
alisonjoseph, line 82: Same thing here

emyarod:
does the user need to set the navType to treeview when there are nested menus or can that be automatically set?

packages/react/src/components/UIShell/components/SideNav.tsx
```diff
@@ -34,6 +34,7 @@ export enum SIDE_NAV_TYPE {
   DEFAULT = 'default',
   RAIL = 'rail',
   PANEL = 'panel',
+  TREEVIEW = 'treeview',
```
emyarod, line 37: I think we briefly discussed this during a dev sync but does the user need to set the navType to treeview when there are nested menus or can that be automatically set?

IgnacioBecerra: For the moment it needs to be automatically set, but we'd have to discuss about the requirements for the automatic setting. Mainly, would the existence of a nested menu mean that the whatever `navType` will be overridden? 

cc: @alisonjoseph @oliviaflory

alisonjoseph:
Just noticed one other thing when I ran the a11y checker on the default SideNav story. Looks like the li still have role="treeitem" and its causing an error.

Screenshot 2025-02-13 at 8 08 25 AM

packages/react/src/components/UIShell/components/SideNav.tsx
emyarod: yeah I think having it tied to the role makes sense and we can revisit this in a future PR :+1:

examples/react/UIShell/src/App.tsx
```diff
@@ -194,8 +194,8 @@ function App() {
             </Theme>
             <Theme theme="g100">
               <SideNav
-                navType={SIDE_NAV_TYPE.PANEL}
                 isChildOfHeader={false}
+                isTreeview={true}
```
alisonjoseph, line 198:
```suggestion
```
I don't think we want this for this secondary nav?

packages/react/src/components/UIShell/__stories__/UIShell.mdx
```diff
@@ -207,6 +202,62 @@ function App() {
 export default App;
 ```
 
+### Treeview
+`SideNav` has a new treeview navigation type that can be enabled by adding the
+`isTreeview` prop. Note that this feature should only be used if there are two
+levels of content within the component.
+
+Also mote that it will be automatically enabled if there are more than two levels of
```
alisonjoseph, line 210:
```suggestion
Also note that it will be automatically enabled if there are more than two levels of
```

packages/react/src/components/UIShell/components/SideNavItems.tsx
```diff
@@ -7,11 +7,13 @@
...
 import cx from 'classnames';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useContext, useEffect, useRef } from 'react';
 import { CARBON_SIDENAV_ITEMS } from './_utils';
 import { usePrefix } from '../internal/usePrefix';
+import { SideNavContext } from './SideNav';
 
 export interface SideNavItemsProps {
+  accessibilityLabel: object;
```
alisonjoseph, line 16: Can we add comments explaining what this is for?

examples/react/UIShell/src/App.tsx
IgnacioBecerra: woops, yeah i'll remove

IgnacioBecerra:
Expected functionality for #415:

https://github.com/user-attachments/assets/10d45070-bfe6-4ec0-af62-ce27bda3844a

Saves menu expanded state so it's back to where you left it if you collapse sidenav and then reopen.

alisonjoseph (approved):
LGTM!

IgnacioBecerra:
@oliviaflory  thanks for the review! We can follow up on those issues in a later PR as they're not related to the tree view features but the Panel view instead. 

For the first issue, I think it would make sense for the previous tab to go to the hamburger menu, but the panel technically part of the page body instead of the menu bar, so it also makes sense for it to go to the last element in the menu...

As for the second issue, I'd say it might be because there's no content, but this one does make sense for the focus to go inside the right panel like in regular carbon.

Maybe we should go over that first issue with Michael Gower? 

Either way, seems like tree view is good design wise, thank you!

cc: @alisonjoseph

oliviaflory:
@IgnacioBecerra oh I see for the first question the Carbon core story behaves the same way, so that makes sense. I didn't realize the main menu panel was part of the page body, thanks for explaining!

oliviaflory (approved):
LGTM 🎉

PR #417, opened 2025-02-05
docs(UIShell): update example

emyarod:
This PR updates the React UIShell sandbox example to match the default story

#### Changelog

**Changed**

- update UIShell exports
- update UIShell sandbox example


alisonjoseph (approved):
thanks, good catch on the missing exports

PR #457, opened 2025-02-20
feat(SideNav): panel flyout menu for SideNavMenu

### LINKED ISSUE #492: [Bug]: SideNavLinks on SideNav Panel should be
### Browser

Chrome

### Package version

latest

### Description

In `SideNav` panel type, the colors of all icons should be `$icon-secondary`, except on hover and selected, where it is `$icon-primary`.

### Screenshots

Image

### Steps to reproduce

Go on SideNav panel story and see the icons' colors.

### Code of Conduct

IgnacioBecerra:
Closes https://github.com/carbon-design-system/carbon/issues/18465 
Closes #492

This PR introduces the FlyoutMenu within the SideNav Panel, where all icons display a tooltip upon either hover or focus. 
The `SideNavFlyoutMenu` component extends the `Tooltip` component, as we needed the regular tooltip to show upon hover, but also needed click functionality. This work also brings three internal functions that are needed for the component to work/

The `SideNavFlyoutMenu` component contains the following interactions:

* Keyboard interactions:
  * Tab into button -> Tooltip with menu title appears
  * Tab into button -> press enter -> menu opens with focus on first element
  * Tab after the last element will make focus go to next button in the SideNav
  * Shift + tab back to the button closes menu and makes tooltip with menu title appear
  * Pressing escape with menu open will close menu and makes tooltip with menu title appear
  * Pressing escape with tooltip open makes tooltip disappear

* Mouse interactions:
  * On hover, the whole menu should appear. If cursor goes away, menu disappears
  * On click, the menu opens and stays open
  * If clicking outside menu, menu will close

#### Changelog

**New**

- `SharkFinIcon` icon
- `SideNavFlyoutMenu` component combining tooltip and popover in one component
- `SideNavMenuItem` has new styles

**Changed**
- `SideNavMenu` renders new flyout menu if parent SideNav is panel type

#### Testing / Reviewing
Test out all the interactions described above in either the `Default` story or `Side Nav Panel With Flyout Menu` story


alisonjoseph:
@IgnacioBecerra I don't see a way to get into the flyout menu items with keyboard. @oliviaflory what is the expected keyboard pattern here. I feel like we didn't want to just have to tab through all the links, but maybe that's what we need to do if it opens on focus/hover? Or does the user tab and then hit enter?

packages/react/src/components/UIShell/components/Sharkfin.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const SharkfinIcon: React.FC<SharkfinIconProps> = ({
+  children,
+  className: customClassName,
+  small = false,
+}) => {
+  const prefix = usePrefix();
+  const className = cx({
+    [`${prefix}--side-nav__icon`]: true,
```
alisonjoseph, line 39: We should probably name this classname the same as the component, and split these styles into their own style sheet. I could see this being completely separate from the sidenav once we bring it to core. Let's also make the file name consistent with the component name.

packages/react/src/components/UIShell/components/styles/_side-nav.scss
```diff
@@ -172,6 +173,75 @@ div:has(.#{$prefix}--header)
...
+      bottom: 4px;
+      right: 4px;
+ 
+      > svg {
+        fill: $icon-secondary;
+      }
+  
+    }  
+  }
+
```
alisonjoseph, line 211: Just a general note that we typically avoid nesting selectors in Carbon.
https://github.com/carbon-design-system/carbon/blob/main/docs/developer-handbook.md#coding-style

oliviaflory:
>Also @oliviaflory, since the SideNavLinkPopover has the "inverse" colors of the theme, I made the flyout match that too. In other words, it's white theme if the side nav panel has the g100 theme. I'm a bit confused since the design specs have them with matching themes, but let me know what's actually expected here.

Yes, it looks a bit funny right now because we're waiting on the ability to customize the tooltip theme, which will come in the next Carbon release.

---

> @IgnacioBecerra I don't see a way to get into the flyout menu items with keyboard. @oliviaflory what is the expected keyboard pattern here. I feel like we didn't want to just have to tab through all the links, but maybe that's what we need to do if it opens on focus/hover? Or does the user tab and then hit enter?

Looks like I didn't write the interaction summary well enough in the parent issue. I've uploaded a revised spec based on Michael's recommendations

Hover
- reveal menu on hover

Focus
- Michael recommended not revealing the menu on focus
- This was the key note I missed writing in the parent issue (sorry!), so let's remove the on focus reveal
- _In this case would we expose a tooltip on focus of the sub-menu name?_ Otherwise if we remove focus for the sub-menus entirely they wouldn't show the user anything.. which might be weird for a sighted keyboard user

Screenshot 2025-02-25 at 3 39 24 PM

OnClick
- We need to show the menu on click in order to support mobile (though we haven't really solidified if the side nav should always remain collapsed on mobile)
- Michael suggested that including an OnClick would also allow the user to navigate past the items with menus if they didn't want to open the menu and be forced to tab through all the links. This is appealing because we don't know how many items will have sub menus in the side nav

He had two options for us regarding keyboard navigation at 41:16 in [the recording](https://ibm-my.sharepoint.com/:v:/r/personal/olivia_flory_ibm_com/Documents/Recordings/UI%20shell%20right%20panel-20250219_150350-Meeting%20Recording.mp4?csf=1&web=1&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D&e=mhWDfN)
1. Tab
   - Spacebar or enter to open menu 
   - Focus moves to first item in menu
   - Tab through menu items, tabbing past the last menu item collapses the menu and focus moves to the next button in the navigation list
   - Esc will collapse the menu and focus will return to the parent item

Screenshot 2025-02-25 at 3 35 51 PM



2. Arrow keys
   - In listening to the recording again we didn't discuss how the menu would open (I assume enter and arrow keys to move around?)
   - He said that the benefit of arrow keys would be that of any menu and if there was ever two levels within the flyout menus arrow key navigation would be better.
  
Screenshot 2025-02-25 at 3 55 12 PM

IgnacioBecerra:
@alisonjoseph @oliviaflory Thanks for the reviews! So at the moment, you do have to hit enter to go into the menu, which would focus into the first link within. It has a focus wrap, so the only way to get out would be to hit enter, or click outside the menu. But looking at Olivia's comment, I'm leaning towards option #1, so I can remove the focus wrap and close after tabbing out of the last item. 

In a later update, we could add a `arrowNav` boolean option to also implement option #2 if needed, so we could have both options.

oliviaflory:
@IgnacioBecerra one more styling fix

on hover of a menu item the text should change to `$text-primary`
Screenshot 2025-02-25 at 5 12 21 PM

alisonjoseph:
@IgnacioBecerra I agree with option 1, I don't think we need to worry about adding a prop for a secondary option.

I'm testing in Firefox in case it makes a difference, but I'm seeing some weirdness with keyboard after tabbing to an icon with a menu (ignore me if you're still working on things)

Enter -> sometimes navigates me directly to the first link, sometimes not
Enter + Tab -> Opens the menu, then tabs to the next icon not inside
Space bar -> Gets me inside the menu, but then I'm stuck and can't tab out of it unless I hit escape. I believe I should be able to tab through to next icon?

IgnacioBecerra:
@oliviaflory FYI when testing the tabbing functionality, there's a weird bug where you can focus into an invisible element before the flyout menu button. This is documented in #483, but I opened a PR to fix that.

oliviaflory:
Thanks Ignacio, this looks awesome and functionality looks good!

I have a few questions from a styling perspective, and you can ignore them if you're still working on things:

1. Is the menu button keyboard focus using the tooltip like the single buttons? They look like two different sizes and we'd want them to be the same. I'm not sure how much it matters whether they both use tooltip vs one tooltip and one popover, but maybe Alison could advise

Screenshot 2025-02-28 at 12 47 31 PM
Screenshot 2025-02-28 at 12 47 57 PM
size comparison


2. Color of the overall tooltip
   - You can't read the OnFocus tooltip, it's a little hard to tell what's happening through inspect. Maybe this will be resolved with the Carbon package update to allow switching the theme
   - the single buttons and flyout menus are opposite colors, I'd expect them to be the same. Could that be caused by popover vs tooltip?

oliviaflory:
Thanks @IgnacioBecerra, this looks awesome and functionality looks good!

I have a few questions from a styling perspective, and you can ignore them if you're still working on things:

1. Is the menu button keyboard focus using the tooltip like the single buttons? They look like two different sizes and we'd want them to be the same. I'm not sure how much it matters whether they both use tooltip vs one tooltip and one popover, but maybe Alison could advise

Screenshot 2025-02-28 at 12 47 31 PM
Screenshot 2025-02-28 at 12 47 57 PM
size comparison


2. Color of the overall tooltip
   - You can't read the OnFocus tooltip, it's a little hard to tell what's happening through inspect. Maybe this will be resolved with the Carbon package update to allow switching the theme
   - the single buttons and flyout menus are opposite colors, I'd expect them to be the same. Could that be caused by popover vs tooltip?

IgnacioBecerra:
@oliviaflory Thanks for checking! Yeah after getting the new style fixes in, the old styles clashed with the changes. I'm fixing the styles. As for the height issue you're seeing, that does come from Popover vs tooltip, but I'll also fix that to match the tooltip properly.

packages/react/src/components/UIShell/__stories__/UIShell.stories.js
```diff
@@ -53,6 +53,7 @@ import {
 } from './AppIcons';
 
 import '../components/ui-shell.scss';
+import { SideNavFlyoutMenu } from '../components/SideNavFlyoutMenu';
```
emyarod, line 56: `'SideNavFlyoutMenu' is defined but never used. Allowed unused vars must match /^_/u`

it looks like this is an unused import?

packages/react/src/components/UIShell/internal/useId.js
```diff
@@ -0,0 +1,117 @@
...
+const useIsomorphicLayoutEffect = canUseDOM ? useLayoutEffect : useEffect;
+
+let serverHandoffCompleted = false;
+
+const defaultId = 'id';
+
+/**
+ * Generate a unique ID for React <=17 with an optional prefix prepended to it.
+ * This is an internal utility, not intended for public usage.
+ * @param {string} [prefix]
```
emyarod, line 56: `Missing JSDoc @param "prefix" description`

just seeing a lint error here for a missing description, that's all

packages/react/src/components/UIShell/internal/useId.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }, []);
+
+  return id;
+}
+
+/**
+ * Generate a unique ID for React >=18 with an optional prefix prepended to it.
+ * This is an internal utility, not intended for public usage.
+ * @param {string} [prefix]
```
emyarod, line 91: another missing description causing a lint error

packages/react/src/components/UIShell/internal/useId.js
```diff
@@ -0,0 +1,117 @@
...
+ * Uses React 18's built-in `useId()` when available, or falls back to a
+ * slightly less performant (requiring a double render) implementation for
+ * earlier React versions.
+ */
+export const useId = _React.useId ? useReactId : useCompatibleId;
+
+/**
+ * Generate a unique id if a given `id` is not provided
+ * This is an internal utility, not intended for public usage.
+ * @param {string|undefined} id
```
emyarod, line 111: another missing description

packages/react/src/components/UIShell/internal/useIdPrefix.ts
```diff
@@ -0,0 +1,16 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+export const IdPrefixContext = React.createContext<string | null | undefined>(
+  null
+);
+
+export function useIdPrefix() {
```
emyarod, line 14: linter is expecting a JSDoc comment for this function

emyarod (commented):
looks good to me, just saw a few things that are causing CI checks to fail!

alisonjoseph:
https://github.com/user-attachments/assets/6e8fca9f-2176-45e0-9421-a191c7621e04

This is looking so good! The only thing I'm seeing now is as issue with when the items don't fit vertically, they squish up instead of scrolling.

IgnacioBecerra:
@alisonjoseph Fixed it in the last commit!

IgnacioBecerra:
@oliviaflory last commit also updates the SideNavLinks to be `$icon-secondary` for the panel icons!

oliviaflory (approved):
styles look good!

packages/react/src/components/UIShell/components/SharkFinIcon.tsx
```diff
@@ -0,0 +1,77 @@
...
+  className?: string;
+
+  /**
+   * Specify whether the icon should be placed in a smaller bounding box
+   * Since the 'small' prop is not provided, we make it optional and set a default value to `false`.
+   */
+  small?: boolean;
+}
+
+export const SharkfinIcon: React.FC<SharkfinIconProps> = ({
```
alisonjoseph, line 32:
```suggestion
export const SharkFinIcon: React.FC<SharkFinIconProps> = ({
```

packages/react/src/components/UIShell/components/SharkFinIcon.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <defs>
+        <clipPath id="clip0_519_52879">
+          <rect width="4" height="4" fill="white" />
+        </clipPath>
+      </defs>
+    </svg>
+  );
+};
+
+SharkfinIcon.propTypes = {
```
alisonjoseph, line 61:
```suggestion
SharkFinIcon.propTypes = {
```

alisonjoseph (changes requested):
Couple nitpicky things with the capitalization in the component name.

Also I think we'll want to rename `shark-fin-icon.scss` to `_shark-fin-icon.scss` and instead of importing directly into the side nav file, import into the root scss file `ui-shell.scss` to keep it as its own component not dependent on side nav. 

Can we also export SharkFinIcon from the index.ts?

One more style thing I noticed, which may be unrelated to this PR, is when the side nav is collapsed the expand button gets hidden if the screen height is too small. It behaves correctly when the side nav panel is expanded.

https://github.com/user-attachments/assets/a6ab393d-25ec-43a2-916d-daf494b9a400

IgnacioBecerra:
@alisonjoseph Address your comments! And seems like the Expand button issue is also happening in the current Storybook:

https://github.com/user-attachments/assets/802e1658-81a1-4810-b7ab-e852a8b57736

I'll open a new issue for this.

alisonjoseph (approved):

PR #465, opened 2025-02-21
feat(uishell): add header story with components update carbon/react

### LINKED ISSUE #395: [Platforms] Implement support for Button and MenuButton inside UI shell header
Design issue https://github.com/carbon-design-system/carbon/issues/18467

### Tasks
- Add ghost `Button` inside `HeaderGlobalBar` to example
- Add `HeaderNavigation` to `example`
- Add two `MenuButton` for tenant switcher inside `HeaderGlobalBar`
- Customize styling for `MenuButton` to match spec
- Extend `HeaderNavigation` with new prop to hide default divider
- New `HeaderDivider` component, similar to `SideNavDivider`
- [ ] Update Stackblitz example

### Design Spec

Image

alisonjoseph:
Closes #395 

#### Changelog

**New**

- add Header example
- add ExpandableSearch to default demo story
- add MenuButton simple category switcher example
- new `HeaderDivider` component
- new styling for MenuButton within Header

**Changed**

- update to latest @carbon/react
- update Header in default story 

### TODO
- [x] Update StackBlitz example


alisonjoseph:
@oliviaflory I dropped several components into the Header and just wanted to confirm which ones we want to support/demo in our storybook. I labeled them all with the component name to make it easier to see which is which. 

https://deploy-preview-465--carbon-labs-react.netlify.app/iframe.html?globals=&args=&id=components-uishell--header-story&viewMode=story

oliviaflory:
Thanks @alisonjoseph! This is great

It looks like the dropdown might be the medium size instead of large so it's slightly mis-aligned with the rest of the components. We'd probably want to do some customizations to adjust the styling to match more similarly to the `HeaderMenu`. Mostly removing the border strong stroke and the background color

I'm have to think a bit more on whether we'd expect a Menu button that would be blue 

I'll also chat with @cameroncalder about how we'd need to unify the selected state between Dropdown, Menu and HeaderMenu, but maybe that would be strictly for Platform guidance vs the UI shell enhancements

alisonjoseph:
@oliviaflory this is ready for another visual review pass
https://deploy-preview-465--carbon-labs-react.netlify.app/?path=/story/components-uishell--default
and
https://deploy-preview-465--carbon-labs-react.netlify.app/?path=/story/components-uishell--header-story

One thing to note, the new HeaderDivider component can be seen on the right side of the screen shot. The divider on the left is included in the HeaderNavigation and has some spacing included. I didn't touch that for now, but let me know if anything needs to get tweaked there.
Screenshot 2025-03-21 at 8 55 22 AM

I wasn't sure the exact order to demo the components in storybook, so let me know if you want things moved around or changed.

oliviaflory:
Thanks Alison, this looks great. I might want to tweak the divider on the left side later but for now we can leave as-is. The order of everything is good!

oliviaflory (approved):
LGTM

IgnacioBecerra (commented):
@alisonjoseph Changes added in this PR look good to me, but I noticed a weird behavior clicking around with the menus:

https://github.com/user-attachments/assets/cbc21f2e-1452-434a-8c4f-f3519792aad2

To reproduce:
1. First click the `MenuButton`
2. Click the `HeaderMenu`, the `MenuButton` closes
3. Click the `MenuButton` again, `HeaderMenu` does not close
4. `HeaderMenu` can't be closed via outside click events but `MenuButton` does. Only upon clicking the menu again will it close

Otherwise, everything's good.

---

### Unrelated to the changes in this PR

I have some questions in regards to accessibility/intended behavior. For what it's worth, these have nothing to do with `carbon-labs`, as the components are coming from `@carbon/react`.

### 1. Focus issue

On the `Default` story, if you're focused on the `MenuButton` (the one with Select Category) and press enter,  the focus goes towards the container with the items, and you either have to tab into it again, or press the Down arrow key to go inside the list. 

On the other hand, in the `Header` story, for some reason after pressing enter, the focus goes straight to the first item.

I left some comments that it might be because the `Header` story is missing the `headerRef`, but maybe having that ref is what is causing this error? Is it necessary to have it?

### 2. Accessibility navigation

Besides that focus issue, once focus is inside, you can only use arrow navigation. Is this intended? It's a bit jarring as the previous menu in the header (`HeaderMenu`) is only navigable with tabs, and both serve relatively the same purpose functionality-wise.


Here's a video of these two issues:

https://github.com/user-attachments/assets/673dcb2f-7ecb-4a47-b83d-ceb66a32cc56


Not sure what a11y navigation standard we want to follow for the Header, but we'd have to either extend one of the components to match the other in our repo, or we'd have to contribute the change to core.

cc: @oliviaflory

alisonjoseph:
@IgnacioBecerra 
I'm looking into the focus issue, I believe the menuTarget/ref is needed to get the correct themeing, but you're right it seems to mess up the focus when using the keyboard. I believe the correct behavior is what you're seeing in the Header story, without the ref. 

> but I noticed a weird behavior clicking around with the menus:

Weird, I'm only able to reproduce this in Chrome, not Firefox (what I use for development). We should probably open an issue in core to track this. 

>Besides that focus issue, once focus is inside, you can only use arrow navigation. Is this intended? It's a bit jarring as the previous menu in the header (HeaderMenu) is only navigable with tabs, and both serve relatively the same purpose functionality-wise.

HeaderMenu is for navigation(which typically uses tabs), and the MenuButton is the tenet switcher, not used for navigating, so different use-case and according to [w3c](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) should use arrow keys. We should probably run this entire thing by Gower/a11y again at some point and get his take on it.

mattrosno:
@alisonjoseph there's something funky going on with the right divider. Sub-pixel aliasing or something. On Chrome at least, this story maybe more:

https://deploy-preview-465--carbon-labs-react.netlify.app/?path=/story/components-uishell--default

The right divider is 0.994px wide when the search box is closed, and if you open the search box, it shrinks to 0.653px wide. Visually, I noticed the divider looking darker when the search box is expanded. You'll also notice the dropdowns slightly shift right as the divider becomes skinner.


https://github.com/user-attachments/assets/2b9cd20b-27d3-485f-923a-1c45501eedb8

alisonjoseph:
@mattrosno super weird, seems to only be happening for me at smaller screen sizes(in FF), I'll take a look. Must be something ExpandableSearch is doing when it expands.

alisonjoseph:
@IgnacioBecerra I opened https://github.com/carbon-design-system/carbon-labs/issues/536 to track the menu focus issues.

PR #466, opened 2025-02-21
fix: remove button style import

alisonjoseph:
This import of Carbon button styles in the example component isn't needed and is overriding UIShell styles due to the order styles are being imported into storybook. 

#### Changelog

**Removed**

- import for button styles in example component


emyarod (approved):
looks good to me!

PR #500, opened 2025-03-06
fix(SideNavToggle): show on screen height resize

### LINKED ISSUE #496: [Bug]: Expand button in SideNavPanel gets hidden on short screen height
### Browser

Chrome

### Package version

latest

### React version

_No response_

### Description

In `SideNavPanel`, when the screen height gets smaller, the expand button on the bottom of the component disappears instead of always remaining at the bottom of the screen. 

Issue can be seen here:

https://github.com/user-attachments/assets/2e79caea-4a3c-4d44-9bb9-686a27632866

### Reproduction/example

see above

### Steps to reproduce

Go to `SideNavPanel` story, and resize the screen height to be smaller. You will notice that at some point the expand button will disappear.

### Suggested Severity

Severity 4 = Unrelated to a user task, has a workaround or does not need a workaround.

### Application/PAL

_No response_

### Code of Conduct

IgnacioBecerra:
Closes #496 

This PR ensures that the toggle button when the side nav panel is collapsed doesn't disappear when the screen height gets smaller.

This is mostly for a visual fix for mouse interaction -- using the keyboard to tab still has a weird behavior. The whole reason this is happening is because of the `overflow` css property that is being applied to the side nav contents. If we remove the values that are currently being applied when the menu is collapsed, **the tooltip/flyout menu will not appear**, as they are considered "overflow" and thus hidden away. To fix this behavior, we would need to refactor most of the menu and also the aforementioned components. For now, this visual fix is a basic compromise to be done.

#### Changelog

**New**

- wrapper div for the toggle button only to ensure black background upon hover

#### Testing / Reviewing
Go to the side nav panel story, and resize the screen height. The toggle button should still appear at the bottom left of the screen no matter what.

### DIFF
--- a/packages/react/src/components/UIShell/components/SideNav.tsx
+++ b/packages/react/src/components/UIShell/components/SideNav.tsx
@@ -508,6 +508,15 @@ function SideNavRenderFunction(
     }
   };
 
+  const SideNavToggleButton = (
+    <SideNavToggle
+      className={!expandedState ? `${prefix}--side-nav__toggle--collapsed` : ''}
+      renderIcon={expandedState ? SidePanelClose : SidePanelOpen}
+      onClick={() => setExpandedState(!expandedState)}>
+      {sideNavToggleText}
+    </SideNavToggle>
+  );
+
   return (
     <SideNavContext.Provider
       value={{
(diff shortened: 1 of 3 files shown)


emyarod (approved):
looks good to me!

PR #512, opened 2025-03-11
fix(UIShell): resolve storybook z-index issue

### LINKED ISSUE #510: [Bug]: Storybook style imports override proper component styles
### Browser

Chrome

### Package version

latest

### React version

_No response_

### Description

**This is not a component bug but rather Storybook specific**

There's an interest bug that is happening specifically on the React Storybook. In the default UIShell story [seen here](https://labs.carbondesignsystem.com/react/?path=/story/components-uishell--default), opening the hamburger menu does not make the SideNav Panel to disappear. 

Image

Looking at the styles for the `.cds--side-nav` element, you can see this:

Image

The styles that are required to make the component work as intended are seen at the very bottom:
```
.cds--side-nav--panel {
    z-index: 7999;
    overflow: visible;
}
```

But they are being overridden by the default `.cds--side-nav` class styles with `z-index: 8000` and `overflow: visible`.

This only happens in production, because of this webpack plugin in the React storybook conf file [seen here.](https://github.com/carbon-design-system/carbon-labs/blob/main/packages/react/.storybook/main.js#L94-100)

To reproduce locally, change `===` to `!==`. It might be happening because for some reason, Storybook is reinjecting the styles, which override the original styles that had already been set (the panel styles, specifically in this case).

Important to note that bug is not present in the general Carbon Labs storybook that contains both WC and React stories (https://labs.carbondesignsystem.com/?path=/story/react_components-uishell--default)

### Reproduction/example

https://labs.carbondesignsystem.com/react/?path=/story/components-uishell--default

### Steps to reproduce

1. Open https://labs.carbondesignsystem.com/react/?path=/story/components-uishell--default
2. Click hamburger menu
3. Side Nav Panel menu should disappear, but remains

### Suggested Severity

Severity 4 = Unrelated to a user task, has a workaround or does not need a workaround.

### Application/PAL

_No response_

### Code of Conduct

IgnacioBecerra:
Closes #510 

This PR fixes the issue where the styles were duplicated throughout all components, meaning that new 'state' styles were overridden by their default styles. This duplication was caused in the latest `WhatsNew` component, where it includes `use @carbon/react;` in the component styles. This makes Storybook grab all the styles from the package and reimport them, essentially.

#### Changelog

**Removed**

- nonessential `@carbon/react` style import

#### Testing / Reviewing
Since this would only change prod Storybook due to the way styles are bundled with Webpack, we can't test if this fixes the issue. 

However, since this change concerns the new What's New component, testing this PR would just be to ensure there are no unexpected changes after doing this change.

### DIFF
--- a/packages/react/src/components/WhatsNew/components/whats-new.scss
+++ b/packages/react/src/components/WhatsNew/components/whats-new.scss
@@ -6,7 +6,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-@use '@carbon/react';
 @use '@carbon/react/scss/zone';
 @use '@carbon/styles/scss/themes';
 @use '@carbon/react/scss/theme' as *;


IgnacioBecerra:
@jlongshore @elycheea This change could affect the WhatsNew component, but there is a higher chance of not having any impact at all. If you could review this just in case, that would be great!

emyarod (approved):
looks good to me!

jlongshore (approved):
LGTM, I see no glaring issues in the Whats new stories at all!

PR #518, opened 2025-03-13
feat(UIShell): add HeaderPopover

### LINKED ISSUE #462: [Platforms] Implement HeaderPanel (right panel) enhancements
Design issue [#18449](https://github.com/carbon-design-system/carbon/issues/18449)

The `HeaderPanel` has already been moved over to Carbon labs - https://labs.carbondesignsystem.com/?path=/story/react_components-uishell--header-panel-story

`HeaderPanel`(s) should open using `HeaderGlobalAction` as a trigger, and be configurable to not be full height.

### Tasks
- Add prop to configure height to match content. (Should there be a min-height?). Maybe autoHeight or hasAutoHeight? 
- Support multiple panels 
- Only one panel open at a time
- Navigation(platform) and panel can not be open at the same time 
- Close when click outside 
- Update Default story
- Address bugs https://github.com/carbon-design-system/carbon-labs/issues/449

### Questions
- Does it need to slide in/out or animate at all?

emyarod:
Closes #462

This PR adds the HeaderPopover component, a popover component used in the header

### DIFF
--- a/packages/react/src/components/UIShell/components/styles/_header.scss
+++ b/packages/react/src/components/UIShell/components/styles/_header.scss
@@ -6,6 +6,7 @@
  */
 @use '@carbon/styles/scss/theme' as *;
 
+@use '@carbon/styles/scss/theme' as *;
 $prefix: 'cds' !default;
 
 .#{$prefix}--header__global {
--- a/packages/react/src/components/UIShell/components/styles/_header.scss
+++ b/packages/react/src/components/UIShell/components/styles/_header.scss
@@ -50,37 +50,3 @@ $prefix: 'cds' !default;
   );
   @include custom-property.declaration('link-focus-text-color', $focus);
 }
-
-.#{$prefix}--header-action.#{$prefix}--animate--slide-down
-  .#{$prefix}--popover
-  .#{$prefix}--header-action__content {
-  transform: translateY(-100%);
-  transition: transform motion(standard, productive) $duration-fast-02,
-    display motion(standard, productive) $duration-fast-02;
-  transition-behavior: allow-discrete;
-
-  /**
-   * starting style also not supported widely
-   * https://caniuse.com/mdn-css_at-rules_starting-style
-   */
-  // stylelint-disable-next-line scss/at-rule-no-unknown
-  @starting-style {
-    transform: translateY(100%);
-  }
-}
-
-.#{$prefix}--header-action.#{$prefix}--animate--slide-down.#{$prefix}--popover--open
-  .#{$prefix}--popover
-  .#{$prefix}--header-action__content {
-  transform: translateY(100%);
-  transition: transform motion(standard, productive) $duration-fast-02;
-
-  /**
-   * starting style also not supported widely
-   * https://caniuse.com/mdn-css_at-rules_starting-style
-   */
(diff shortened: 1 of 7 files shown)


emyarod:
If you go to the `Header` story, change the `animate` storybook control to `SLIDE_DOWN`, and open the HeaderAction panel ("i" icon in the top right), you can see the menu also slides over the header. At the moment I don't think we can resolve that unless we change the nesting structure of the core header. Still investigating

emyarod:
the link in the popover should now be the correct color, and the other header item tooltips will be resolved in a separate PR (awaiting a Carbon version upgrade)

alisonjoseph (changes requested):
This looks great! I only have a couple small comments. 

Can you double check the svg icon fill color inside the HeaderPopoverButton, I think it should be `icon-secondary` and then `icon-primary` on hover? cc @oliviaflory 

Can you update the package readme, the example and the example code in the mdx story file?

When you fix the merge conflicts can you add this to the Header story also?

IgnacioBecerra (approved):
LGTM

oliviaflory:
@emyarod alison's comment is correct, we want to the svg to be $icon-secondary, and then $icon-primary on hover + while it's open. Looks like the 9dot switcher is working similarly.

packages/react/src/components/UIShell/components/styles/_header.scss
```diff
@@ -5,8 +5,35 @@
...
+  @include custom-property.declaration(
+    'link-hover-text-color',
+    $link-primary-hover
+  );
+  @include custom-property.declaration(
+    'link-visited-text-color',
+    $link-visited
+  );
+  @include custom-property.declaration('link-focus-text-color', $focus);
+}
```
alisonjoseph, line 39:
```
@include custom-property.declaration('button-focus-color', theme.$focus);
  @include custom-property.declaration('link-focus-text-color', theme.$focus);
```

I think we need to add these two tokens also to make sure we're swapping back everything added here: https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/toggletip/_toggletip.scss#L53-L74

oliviaflory:
@emyarod thanks for addressing the comments I left for hover and the focus state inside the popover.

Have you noticed that the other buttons in the header look like they have a 3px focus state now? Only the buttons with the blank checkbox are thicker than the other elements within the shell

Screenshot 2025-04-01 at 12 52 32 PM
Screenshot 2025-04-01 at 12 56 57 PM

alisonjoseph (commented):
Can you add the new components to the subcomponents list at the top of the stories file, this way we get the component api on the Overview page. 

Might also be good to add `tooltipHighContrast={false}` to the `HeaderGlobalAction` components in the `Header` story so the tooltips all match.

alisonjoseph (approved):

alisonjoseph:
@oliviaflory I think this is good to merge, but just wanted to double check with you for approval first. I opened an [issue](https://github.com/carbon-design-system/carbon-labs/issues/551) for the 3px focus state on the HeaderGlobalAction buttons.

oliviaflory (approved):
Thanks @emyarod 🎉

PR #523, opened 2025-03-17
fix(animated-header): bug fixes

glapadre:
Closes #

Add `welcomeText` prop to deal with internationalization issue of static "Welcome" text
Fix bug related to dropdown button overflow use in Animated header
AI tile and Glass tile same height
Update snapshot

#### Changelog

**New**

- N/A

**Changed**

-Note above

**Removed**

- N/A

#### Testing / Reviewing

Test with storybook


packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
```diff
@@ -258,7 +259,9 @@ const AnimatedHeader: React.FC<AnimatedHeaderProps> = ({
                   className={`${blockClass}__workspace`}
                   size="sm"
                   titleText="Label"
-                  label={`Open in: ${name}'s workspace` || 'Select a workspace'}
+                  label={
+                    `Open in: ${userName}'s workspace` || 'Select a workspace'
+                  }
```
IronZDev, line 264: What do you think about exporting this label as prop as well? It will solve the internationalisation problems for it.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
```diff
@@ -231,7 +233,6 @@ const AnimatedHeader: React.FC<AnimatedHeaderProps> = ({
                   id={`${blockClass}__header-dropdown`}
                   className={`${blockClass}__header-dropdown`}
                   size="md"
-                  autoAlign
                   titleText="Label"
                   initialSelectedItem={allTiles[0]}
                   label={allTiles[0].name || 'Select an option below'}
```
IronZDev, line 238: What do you think about exporting this label as prop as well? It will solve the internationalisation problems for it and give move flexibility. We have a requirement to have a custom label non-dependant on first tile name.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
```diff
@@ -165,7 +167,7 @@ const AnimatedHeader: React.FC<AnimatedHeaderProps> = ({
       // Run Looped Animation
       animation.addEventListener('complete', loop);
     }
-  }, []);
+  }, [animationContainer, headerAnimation, isReduced]);
```
IronZDev, line 170: I believe we should add a cleanup function here to make sure there are no dangling eventListeners.
```suggestion
    return () => {
      animation?.removeEventListener('DOMLoaded', reducedMotion);
      animation?.removeEventListener('DOMLoaded', load);
      animation?.removeEventListener('complete', loop);
    } 
  }, [animationContainer, headerAnimation, isReduced]);
```

IronZDev (commented):
Nice job! I have added few additional suggestions, please take a look.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
glapadre: No issue with doing this to fix the internationalization issue, although the designs recommend starting with a similar text format of `Open in: ${userName}'s workspace` depending on the language. Once a item is clicked from the dropdown it is removed essentially acting as just the placeholder.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
glapadre: Can you show an example where the label wouldn't represent the Tile Group that is selected? 

I just want to understand logically how this would work. In the current setup, the Dropdown label represents the tiles/items selected to the right. 

Since the labels are coming from the Tile data there should be no issues with internationalization since the data is directly coming from the end user. The 'Select an option below' is the backup option which is English only currently.

mikeolasov (approved):

PR #490, opened 2025-03-04
feat(processing): add new processing component

glapadre:
Closes #

New shared processing component. Currently being used across design teams to show processing data or content in AI Chat shell. Used with the IBM Automation team right now.

#### Changelog

**New**

This is a new component so everything in packages/react/src/components/Processing

**Changed**

- N/A

**Removed**

- N/A

#### Testing / Reviewing

Test with storybook

PR #533, opened 2025-03-26
fix: duplicate style import storybook bug

alisonjoseph:
#490  was merged in before the issue here could be fixed https://github.com/carbon-design-system/carbon-labs/pull/490#pullrequestreview-2718182247

### Testing
Check that the UI Shell and Processing components both render correctly with this import removed.


glapadre:
Covered in #534

PR #539, opened 2025-03-27
fix: correct package exports

lee-chase:
Importing the UI Shell directly from the package is not currently possible

e.g.
`import { SideNav } from '@carbon-labs/react-ui-shell';` fails as it directs to `./src/index.js` which is not exported.

The package exports for all main and module currently point a the source folder.

#### Changelog

**Changed**

In all package.json

```
  "main": "./src/index.js",
  "module": "./src/index.js",
```

to

```
  "main": "lib/index.js",
  "module": "es/index.js",
```

#### Testing / Reviewing

N/A


lee-chase:
Hmm based on this perhaps they should switch to exports. 

https://nodejs.org/api/packages.html#conditional-exports

NOTE: Not touched web components or utilities which have similar issues. Perhaps decide on best format first.

lee-chase:
It looks like the correct form is defined here https://nodejs.org/api/packages.html#conditional-exports

No `main` or `module` but `exports`

```
  "exports": {
    "import": "./index-module.js",
    "require": "./index-require.cjs"
  },
```

tay1orjones (approved):
This being labs I'd say use the most current approach w/ `exports`. It sounds like you might want [subpath](https://nodejs.org/api/packages.html#subpath-imports:~:text=the%20exports%20field.-,Subpath%20patterns,-%23) exports as well. 

The monorepo doesn't do this yet because old (versions of) bundlers, like webpack 4, do not handle `exports` correctly and it's a bit of a minefield to get the best support.

[This is a great writeup](https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/) from 2023 and honestly not much has changed at all since then 😓

annawen1 (approved):
I tested with Web Components packages since some of them are actively being used and they aren't affected, so I'm good with these changes!

PR #545, opened 2025-04-01
feat(UIShell): add TrialCountdown component

alisonjoseph:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/packages/react/src/components/UIShell/__stories__/UIShell.mdx
+++ b/packages/react/src/components/UIShell/__stories__/UIShell.mdx
@@ -43,6 +43,7 @@ package:
 - `HeaderPanel`
 - `HeaderDivider`
 - `SharkFinIcon`
+- `TrialCountdown`
 
 <Canvas of={UIShellStories.Default} />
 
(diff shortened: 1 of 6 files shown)


alisonjoseph:
@oliviaflory I was matching what was there for MCSP, but I agree it feels a little tight. They have a min-width set to 16px, so when there is only one number in the countdown it looks better.
Screenshot 2025-04-07 at 8 12 56 AM

emyarod (commented):
just one thing I noticed about the focus border on the lines, it seems they are full width but I'm not sure what the expected behavior should be
image

alisonjoseph:
@emyarod its because .`cds--toggletip-content` has `display:grid` set. Not sure that we want to override that, or do some custom styles for the countdowb popover.

emyarod:
> @emyarod its because .`cds--toggletip-content` has `display:grid` set. Not sure that we want to override that, or do some custom styles for the countdowb popover.

ah yeah actually I think there usually is a wrapper flexbox div around the button set, but it makes the links inline-flex so they end up being on the same line. but there wasn't enough space with the icons so I removed the div. this may be fine as is!

PR #554, opened 2025-04-04
fix(UIShell): remove box-shadow on header action focus

### LINKED ISSUE #551: [Bug]: HeaderGlobalAction has double focus states
Image

Focus state is showing up 3px on `HeaderGlobalAction`, looks like this is due to it getting both a border and a box shadow on focus. Should be 2px

emyarod:
Closes #551

This PR removes the box shadow on header global actions so that the focus state is no longer doubled

#### Changelog

**Removed**

- header global action box shadow

#### Testing / Reviewing

confirm the component focus state looks correct


alisonjoseph (approved):
LGTM!

PR #560, opened 2025-04-08
chore: add toolbar example as a pattern (story only)

devadula-nandan:
Closes https://github.com/carbon-design-system/ibm-products/issues/7111

adds an example code to show how we can build toolbar out of carbon components
adds stackblitz examples

#### Changelog

**New**

- added story for toolbar (story only, there is no component)
- added stackblitz example


packages/web-components/src/patterns/toolbar/__stories__/toolbar.mdx
```diff
@@ -0,0 +1,296 @@
+import { ArgTypes, Canvas, Markdown, Meta } from '@storybook/blocks';
+import { cdnJs, cdnCss } from '../../../globals/internal/storybook-cdn';
+import * as ToolbarStories from './toolbar.stories';
+
+<Meta of={ToolbarStories} />
+
+# Toolbar
+
+- **Initiative owner(s):** Nandan Devadula
```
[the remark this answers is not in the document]
devadula-nandan: oops my bad 😁. thats an already existing pattern. thanks for clarifying.

examples/web-components/toolbar/package-lock.json
elycheea, line 1: Let’s remove the `package-lock` since we don’t need to include these in examples.

devadula-nandan: @elycheea, i had to refactor a lot of this structure, to be consistent with how we are delivering patterns. taking reference from ibm products.
also adds a complete lit example instead html, being consistent with ibm-products
i also had to add a lit config to load lit example styles in storybook.

pls have a second look when you get some time. 😄

PR #579, opened 2025-04-28
Feat/resize bar

devadula-nandan:
Closes #

resize bar draft

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

#### Testing / Reviewing

### DIFF
--- a/examples/react/ResizeBar/index.html
+++ b/examples/react/ResizeBar/index.html
@@ -1,0 +1,22 @@
+<!--
+@license
+
+Copyright IBM Corp. 2025
+
+This source code is licensed under the Apache-2.0 license found in the
+LICENSE file in the root directory of this source tree.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>@carbon-labs/react-resize-bar stackblitz</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/main.tsx"></script>
+  </body>
+</html>
--- a/examples/react/ResizeBar/package.json
+++ b/examples/react/ResizeBar/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "carbon-labs-react-resize-bar-example",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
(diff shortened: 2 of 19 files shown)


elycheea:
During sizing/refinement pt 2 today, we were leaning towards the name `Resizer` instead of `ResizeBar` since that feels a bit more generic and allows for more flexibility if there are future iterations on it.

packages/react/src/components/Resizer/__stories__/Resizer.stories.js
```diff
@@ -0,0 +1,538 @@
...
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useRef } from 'react';
+import mdx from './Resizer.mdx';
+import { Resizer } from '../components/Resizer';
+import '../components/resizer.scss';
```
devadula-nandan, line 14: note to self: move this to resizer

lee-chase:
In general the component works well and represents the intent of the design. 

There is a stolen focus issue that means the drag line can jump or be moved without the mouse being depressed. 

- Right click on the drag line to show the context menu on the single panel no boundaries story and select 'inspect' and note the line jumps. It may be possible to view this in other scenarios. 
- With another application open, right click on the line to show the context menu and then use Cmd+Tab to switch to the other application. When you click back to the browser the line is draggable even though the mouse button has been released.
- The implementation uses non-context layer tokens such as layer-01 and border-subtle-01. In general we should avoid using  non-contextual values (those with numbers) in favour of the layer context sensitive versions e.g. layer and border-subtle.

NOTE: Personally I find the visuals lack a clear drag handle and the thickened line does unintuitive. This is especially true if the drag line meets a screen edge (see image). I have mocked up a small drag handle below. This could be used on both sides with multiple panels.

image
image
image

lee-chase:
Looks good @devadula-nandan and I will approve.

I think things to explore are:
- A persistent drag handle that is reminiscent of the text area resizer.
- Consolidating with SplitPanel. The main difference I see is that split panel permits a gap between panels based on the grid gutter options.

Perhaps a small working group or dsag discussion.

lee-chase (approved):
LGTM should consolidate with split panel and iterate on drag handles.

devadula-nandan:
> Looks good @devadula-nandan and I will approve.
> 
> I think things to explore are:
> 
> * A persistent drag handle that is reminiscent of the text area resizer. Personally like the last custom example, but would leave the handle permanently visible, but only on the sides where panels are attached to the resizer.
> * Consolidating with SplitPanel. The main difference I see is that split panel permits a gap between panels based on the grid gutter options.
> 
> Perhaps a small working group or dsag discussion.

added few more examples, which i believe are good to have. we can make any of the decided handle or any new recommendation as part of the component, until we get any approval from the design, till then i think we can leave it as an open canvas.

we can add the gutter manually mentioned in docs
> A gap can be added to the parent if the design demands.

https://github.com/user-attachments/assets/e5417958-ae26-4e6c-9d5f-1b2733b4514f

PR #605, opened 2025-05-20
fix: update exports to match UI shell

lee-chase:
Related to #603 

Given the verified example of fixing the UI Shell exports and example application, this PR update the remaining react package exports and example imports.

Slight change to sub folder exports to match https://nodejs.org/api/packages.html#subpath-exports

#### Changelog

**Changed**

- React package.json files to match UiShell
- Example imports to remove es/index.js
- Updated generate script file.

#### Testing / Reviewing

Was verified locally by checking/updating examples.
To be verified by opening stackblitz examples after publish.

### DIFF
--- a/examples/react/AnimatedHeader/src/App.tsx
+++ b/examples/react/AnimatedHeader/src/App.tsx
@@ -8,16 +8,21 @@
  */
 
 import React, { useState } from 'react';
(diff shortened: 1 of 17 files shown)


alisonjoseph (approved):

alisonjoseph:
@lee-chase this seems to be breaking things, can't run storybook locally. Any ideas or should I revert for now?

Issue #610, opened 2025-05-22
[Platforms]: SideNav - Remove rail tooltips and flyout menu


alisonjoseph:
The entire rail will now open on hover so there is no need for the tooltips or popovers

![Image](https://github.com/user-attachments/assets/fba0f869-3012-4522-8f8b-5b766ad7545f)

![Image](https://github.com/user-attachments/assets/c842b937-c28c-4a00-b277-86b4ec29181a)

Here is PR where it was added https://github.com/carbon-design-system/carbon-labs/pull/457

### Tasks
- When navType={SIDE_NAV_TYPE.PANEL} remove tooltip and popovers 

### Questions

- Is there any variant of the shell where we would need to leave this in?

alisonjoseph:
@mattrosno moving this to the enhancements milestone as it doesn't affect anything we're doing for June 15th.

alisonjoseph:
HOLD on this - don't remove yet

alisonjoseph:
closing, we are keeping this functionality

Issue #593, opened 2025-05-13
[Platforms]: UI Shell SideNav - Phase 2


alisonjoseph:
Includes
- double wide
- 5 levels of nesting
- updated keyboard functionality
- mobile sie updates
- slot for filtering or dropdown
- rail updates

PR #617, opened 2025-05-27
feat(uishell): update to use rail instead of panel nav remove inline theme

### LINKED ISSUE #611: [Platforms]: SideNav - Add ability to customize icon for HeaderMenuButton
`HeaderMenuButton` needs to be able to render a switcher icon

### Tasks
- Copy `HeaderMenuButton` over from `@carbon/react`
- Add renderIcon prop

### Questions
- Do we want to allow any icon with a renderIcon prop, or do we want to limit it to the hamburger or switcher icon?

alisonjoseph:
Ref https://github.com/carbon-design-system/carbon-labs/issues/593
Closes #611 
Ref #610 (❓Do we need to remove this from labs if we're not using it) 
Ref #618

This PR updates our examples to be as close as possible to updated designs without adding anything new.

This removes inline theming, so everything is the same theme, and also takes the new panel nav type out of the example and uses the isRail prop from core to open the side nav on hover.

Updates HeaderMenuButton to use Switcher (9 dot) icon

This also removes the Switcher from the examples.  

#### Changelog

**New**

- SideNav with Rail story

**Changed**

- Update z-index for isRail nav to show up behind header
- Update default story & Stackblitz to use isRail prop instead of navType=panel
- Update default story & Stackblitz to use 9 dot switcher icon

**Removed**

- Remove inline theming from default story & Stackblitz
- Remove switcher from default story & Stackblitz
- Remove Header story


ariellalgilmore (approved):

emyarod (approved):
looks good to me!

PR #620, opened 2025-05-28
feat(uishell): update navigation to be overlay and not grid influencer

### LINKED ISSUE #619: [Platforms]: SideNav - add option for floating nav (default is grid influencer)
The default SideNav behavior, when paired with the Content component, is a grid influencer. When the navigation is collapsible at desktop or rendered as a rail, it should instead behave as a floating panel. This means it should overlay the content rather than shift it. The overlay should also display at all breakpoints when the navigation is visible.

### Tasks
- [ ] When `hideOverlay` is false SideNav should be floating not a grid influencer.
- [ ] Update overlay to show at all breakpoints when `hideOverlay` is false
- [ ] Add functionality to close/collapse `SideNav` on click of overlay

alisonjoseph:
Closes #619 

This PR switches both SideNavs (primary and local) to be overlays and not grid influencers and add overlay. 

#### Changelog

**New**

- add new overlay class to SideNav when hideOverlay prop is false
- add styling to ensure its not a grid influencer when there is an overlay
- update styles to ensure overlay can display at all breakpoints

**Changed**

- update StackBlitz example
- update snapshots to fix failing tests

### DIFF
--- a/examples/react/AnimatedHeader/src/App.tsx
+++ b/examples/react/AnimatedHeader/src/App.tsx
@@ -8,16 +8,21 @@
  */
 
 import React, { useState } from 'react';
-import { AnimatedHeader } from '@carbon-labs/react-animated-header/es/index';
 import {
+  AnimatedHeader,
   watsonXAnimatedLight,
   watsonXStaticLight,
-} from '@carbon-labs/react-animated-header/assets';
-import { headerTiles, workspaceData, tasksConfigDropdown } from './data';
+} from '@carbon-labs/react-animated-header';
+
+import {
+  headerTiles,
+  workspaceData,
+  tasksConfigDropdown,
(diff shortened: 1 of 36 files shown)


ariellalgilmore (approved):
LGTM! just added a comment to use a spacing token

emyarod (approved):
looks good to me!

oliviaflory (approved):
LGTM!

The only funky thing I'm noticing is on small screens when you click one of the rail items the siderail briefly expands and then navigates to the new page. 

https://github.com/user-attachments/assets/24b9495b-08ae-4950-a73e-aebb590e4830


Happy to tweak this later, I was thinking the rail would remain collapsed, but we're also hiding it on mobile so that probably doesn't matter!

PR #624, opened 2025-05-30
fix: add breakpoint utility for responsive side-nav rail

emyarod:
Related #618

this PR adds a media query to hide the side navigation rail on smaller screens

#### Changelog

**Changed**

- sidenav rail styles

#### Testing / Reviewing

confirm that the local sidenav rail is not visible at smaller screen sizes

### DIFF
--- a/packages/react/src/components/UIShell/components/SideNav.tsx
+++ b/packages/react/src/components/UIShell/components/SideNav.tsx
@@ -62,6 +62,7 @@ export interface SideNavProps
     event: FocusEvent<HTMLElement> | KeyboardEvent<HTMLElement> | boolean,
     value: boolean
   ) => void;
+  hideOnBreakpoint?: 'sm' | 'md' | 'lg' | 'xlg' | 'max';
   href?: string;
   isFixedNav?: boolean;
   isRail?: boolean;
@@ -100,6 +101,7 @@ function SideNavRenderFunction(
     children,
     onToggle,
     className: customClassName,
+    hideOnBreakpoint,
     href,
     isFixedNav = false,
     isRail,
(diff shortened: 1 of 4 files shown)


alisonjoseph (approved):

alisonjoseph (changes requested):
I approved too quickly, I think we want the rail visible at the medium breakpoint.

alisonjoseph (approved):

ariellalgilmore (approved):

alisonjoseph (changes requested):
Nice, one little thing, I think we want the default to not do anything, so will want to add the prop to our default story (and stackblitz example) to hide at the small breakpoint.

packages/react/src/components/UIShell/components/SideNav.tsx
```diff
@@ -599,6 +602,12 @@ SideNav.propTypes = {
    */
   expanded: PropTypes.bool,
 
+  /**
+   * Specify the breakpoint at which the SideNav will be hidden.
+   * Can be one of `sm`, `md`, `lg`, `xlg`, or `max`.
+   */
+  hideOnBreakpoint: PropTypes.oneOf(['sm', 'md', 'lg', 'xlg', 'max']),
```
alisonjoseph, line 609: We should probably add an additional comment line

`* Only applies when `isRail` is true.`

alisonjoseph (approved):
LGTM!

ariellalgilmore (approved):
LGTM!

aubrey-oneal:
Looks good based on my review of the [deploy preview](https://github.com/orgs/carbon-design-system/projects/90/views/1)

PR #628, opened 2025-05-30
feat(ui-shell): double wide

### LINKED ISSUE #596: [Platforms]: SideNav - Implement double wide nav
> [!CAUTION]
> Needs design spec 

### Questions

ariellalgilmore:
Closes #596 

implement double wide

#### Changelog

**New**

- this adds a new primary prop to the SideNavMenu
- if primary SideNavMenu is being used it will render children inside SideNavItems
- the "secondary" panel should never close
- on mobile it should hide and show primary/secondary one at a time and there is a back button with props to allow for user inputted icon and text
- new currentPrimaryMenu on SideNavContext to keep track on which primary menu is open
- keyboard nav updates
- and some style changes

#### Testing / Reviewing

check both desktop and mobile variations

### DIFF
--- a/packages/react/src/components/UIShell/components/SideNavMenu.tsx
+++ b/packages/react/src/components/UIShell/components/SideNavMenu.tsx
@@ -330,8 +330,9 @@ export const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
               if (onMenuToggle) {
                 onMenuToggle();
               }
-              setIsExpanded(false);
-
+              if(!primary && isExpanded){
+                setIsExpanded(false);
+              }
               // go to previous level's side nav menu button
             } else {
               // since we're in a menu, it finds its own <li>, we go up one more
@@ -387,6 +388,8 @@ export const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
     useEffect(() => {
       if (currentPrimaryMenu !== uniqueId) {
         setIsExpanded(false);
+      } else {
+        setIsExpanded(true)
       }
     }, [currentPrimaryMenu]);
 
@@ -414,7 +417,8 @@ export const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
         aria-expanded={isExpanded}
         className={className}
         ref={listRef}
-        onKeyDown={handleKeyDown}>
+        onKeyDown={handleKeyDown}
+        id={uniqueId}>
         <button
           aria-expanded={isExpanded}
           className={buttonClassName}
--- a/packages/react/src/components/UIShell/components/SideNavMenu.tsx
+++ b/packages/react/src/components/UIShell/components/SideNavMenu.tsx
@@ -24,6 +24,7 @@ import React, {
   useState,
 } from 'react';
 import { CARBON_SIDENAV_ITEMS } from './_utils';
+import { useId } from '../internal/useId';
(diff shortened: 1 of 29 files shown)


packages/react/src/components/UIShell/components/SideNavMenu.tsx
```diff
@@ -73,6 +94,11 @@ export interface SideNavMenuProps {
    */
   isSideNavExpanded?: boolean;
 
+  /**
+   *  Specifices if it's the primary sidenav
```
alisonjoseph, line 98:
```suggestion
   * Specifies if this is the primary SideNav.
   * If true, child components will open to the right,
   * creating the double-wide navigation layout
```

Adding some more comments here

packages/react/src/components/UIShell/components/SideNavMenu.tsx
```diff
@@ -18,14 +24,29 @@ import React, {
...
+const smMediaQuery = `(max-width: ${breakpoints.md.width})`;
+
 export interface SideNavMenuProps {
+  /**
+   * Title for back button in sm screen
+   */
+  backButtonTitle?: string;
+
+  /**
+   * A custom icon to render on the back button in sm screen
```
alisonjoseph, line 46:
```suggestion
   * A custom icon to render on the back button in sm screen
   * default is ArrowLeft
```

ariellalgilmore:
thanks @alisonjoseph!! I updated the active and button styles, but thinking maybe the nested level spacing can be a separate issue

alisonjoseph:
Just adding this here so we don't forget before merging
- [ ] Update Stackblitz code in examples folder
- [ ] Update Overview.mdx page to add documentation around double-wide implementation

oliviaflory:
### Desktop

- [ ] 1. Small bug when I press space on the sub menu within panel 2 it flashes like it's trying to open but remains closed. Should space do nothing or open the menu? It looks like it opens/selects the menu from panel 1 (tested on Chrome).

https://github.com/user-attachments/assets/06f9b95b-2be6-4918-ab40-1bb52532397f

---

- [ ] 2. Tabbing in Firefox. It's a bit unclear to me where the focus is going (within the panels somewhere?) when I hit the `tab` key on Firefox. Two examples in the video, and also attaching a screen shot of the odd focus indicator that seems like it's between panels?

Screenshot 2025-06-06 at 2 18 12 PM

https://github.com/user-attachments/assets/eb54f362-a379-4aa3-a577-e08a26c375b6

---

- [ ] 3. First tab stop when opening the menu. This might be because our storybook isn't a real environment but want to double check expected behavior.
I was expecting the focus to move to the 'selected or active' page on the first tab stop (item 3) but right now the first tab stop is Product 1. Is that a realistic expectation/behavior for a tree? I thought that is what Gower had told us in a previous meeting but that may have been when he thought it wasn't a tree and everything was tab stops.

---

- [ ] 4. Is it possible for the 'selected or active' page parent to remain open when browsing other product menus? Again this might be because storybook isn't a real product environment.
Right now we see Home expanded and Item level 3 selected upon opening the menu. If I click to a different Product (2 or 4) to view my options, and then go back to Product 1, I would think Home would still be expanded but it us collapsed and I have to open to see which page I'm already on. Is that a reasonable behavior to expect?


https://github.com/user-attachments/assets/3432b7ed-bd5e-4fca-a420-96ed35324b8e

Issue #601, opened 2025-05-19
[Feature Request]: Enhance the UI Shell example to be more realistic


lee-chase:
### The problem

The UI Shell is currently built with typescript, but has a whole bunch of typescript issues.

- Often complaints regarding types being any
- There is a lint script, but no eslint or config
- There is no actual navigation
- There is not testing - at a minimum the project should lint and build without raising errors.
- Three is not linting of style which could hide issues the end user will see such as 

![Image](https://github.com/user-attachments/assets/62ffa8c1-25a4-4f97-a1c5-36c565f9c9a1)

### The solution

- Ensuring typescript is used correctly in the example.
- Add eslint dev dependency and basic script
- Add stylelint dev depended and basic script
- Add react-router and some actual navigation
- Add a smoke test to verify linting and build function as expected with each version.

### Business priority

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-labs/blob/main/.github/CODE_OF_CONDUCT.md)

PR #648, opened 2025-06-09
feat: add React router to UIShell example

lee-chase:
Contributes to https://github.com/carbon-design-system/carbon-labs/issues/601

The UI shell example has no routing. This PR adds react-router as a more realistic exammple.

#### Testing / Reviewing

Built/ran locally.


lee-chase:
@alisonjoseph & @ariellalgilmore as per slack chat. I think primary and defaultExpanded can be added to config. For SideNavSlot I need to understand if this can contain routes, if so how they're used.

ariellalgilmore:
hey, @lee-chase! I created a [stackblitz](https://stackblitz.com/github/ariellalgilmore/ui-shell-demo?file=src%2FApp.tsx&tab=readme-ov-file) with how the react routing could possibly be added with what the new features that have been implemented in the UIShell. 

Right now you can click on the following links:
Screenshot 2025-06-20 at 2 43 54 PM

and the `isActive` page will update only on the SideNav, but should be simple to update this to the local rail as well. 

Let me know if you wanna jump on call to discuss it more. I know adopters are excited to have this example!

lee-chase:
Hi @ariellalgilmore 

Not a fan of lumping everything into the App as tends to be the case with Carbon examples. In IBM products the examples were in there own files. Which has the benefit you could add multiple versions

E.g. https://stackblitz.com/edit/github-c2q9waig-dkyemrc4?file=src%2Fcomponents%2FHeaderExample.tsx

lee-chase:
Hey @ariellalgilmore as per slack discussion, here is another [stackblitz](https://stackblitz.com/edit/github-rpwun6bn).
FYI @alisonjoseph 

I have also updated this PR.

It does
- Include platform and product level navigation
- Includes react-router just for product 1 side nav.

It does NOT:
- Extract the platform navigation structure to config or react-router

I did not want it to make further updates without fully understanding the consequences of accessing another product. For example could this change the global navigation? Is the platform the owner of the overall navigation? 

Whether going with this version or the non-ReactRouter version I would recommend switching away from a single file example.  Splitting out the global bar,, side nav and story content makes it much easier to parse.

Perhaps we can have a chat later in the week.

P.S. Platform sounds nicer than primary to me.

ariellalgilmore:
Hi lee! messaged on slack but this looks awesome!! for the platform navigation I think it would be awesome if every link had real navigational actions like the first platform. Each platform change would have to update the "local" or "SideNav isRail"  to match the link scheme in the "primary" or "product" SidePanel.

lee-chase:
> stackblitz

Agree.

alisonjoseph:
Anyone else seeing this in the example but not in storybook?
Screenshot 2025-07-15 at 1 16 30 PM

ariellalgilmore:
@alisonjoseph hmm not me
Screenshot 2025-07-15 at 11 43 53 AM

ariellalgilmore (approved):

lee-chase:
@alisonjoseph Close but there was a repeat render issue in SideNavMenu and I haven't had a chance to circle back. Going to be mid August now as I'm on vacation from tomorrow.

image

PR #651, opened 2025-06-09
fix(ui-shell): tabbing fixes

ariellalgilmore:
Ref #649 

>  Tabbing in Firefox. It's a bit unclear to me where the focus is going (within the panels somewhere?) when I hit the tab key on Firefox. Two examples in the video, and also attaching a screen shot of the odd focus indicator that seems like it's between panels?

^there shouldn't be any weird tabbing anymore in any browser

#### Changelog

**Changed**

- edited so if it's a primary menu it renders one section of code or the other and not both 
- set tabIndex = -1 if it's a primary menu
- ignores setting a link in a primary sidenavmenu to tabindex = -1 

#### Testing / Reviewing

tab through the sidenav and make sure in firefox that you aren't seeing any random blue lines

### DIFF
--- a/packages/react/src/components/UIShell/components/SideNav.tsx
+++ b/packages/react/src/components/UIShell/components/SideNav.tsx
@@ -547,7 +547,11 @@ function SideNavRenderFunction(
     items.forEach((item) => {
       if (
         item.classList.contains(`${prefix}--side-nav__toggle`) ||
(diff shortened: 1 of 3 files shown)


aubrey-oneal (approved):
Looks good, thanks for jumping on a call!

PR #653, opened 2025-06-10
fix(sidenav): focus on active

### LINKED ISSUE #649: [Platforms]: SideNav - react tabbing and focus issues

ariellalgilmore:
Closes #649 

if there is an active link then focus should go to that item first on load

- {{new thing}}

**Changed**

- update focus on first item if there is an active link
- add an overflow to the seconary menu
- make sure escape key is handled
- commented out the `lastExpandedState` seemed to be causing errors and not doing much from what I can tell

#### Testing / Reviewing

check on load tabbing

### DIFF
--- a/packages/react/src/components/UIShell/components/SideNavMenu.tsx
+++ b/packages/react/src/components/UIShell/components/SideNavMenu.tsx
@@ -439,6 +439,11 @@ export const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
       } else {
         // setIsExpanded(lastExpandedState);
       }
+
+      // will always open to the menu with an active element
+      if(primary && (active || hasActiveDescendant(children))){
+        setIsExpanded(true)
+      }
     }, [sideNavExpanded]);
 
     const [openPopover, setOpenPopover] = React.useState(false);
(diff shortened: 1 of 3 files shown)


packages/react/src/components/UIShell/components/SideNavMenu.tsx
```diff
@@ -421,23 +421,23 @@ export const SideNavMenu = React.forwardRef<HTMLElement, SideNavMenuProps>(
...
+      if (primary && currentPrimaryMenu !== uniqueId) {
         setIsExpanded(false);
       } else {
         setIsExpanded(true);
       }
     }, [currentPrimaryMenu]);
 
     // save expanded state before SideNav collapse
-    const [lastExpandedState, setLastExpandedState] = useState(isExpanded);
+    // const [lastExpandedState, setLastExpandedState] = useState(isExpanded); // this seems to be causing issues and not really needed?
```
alisonjoseph, line 432: Wanted to just check in with @IgnacioBecerra before we remove this.

alisonjoseph (commented):
Overall looks good to me, wanted to check in with Ignacio before we totally remove that code, but I think should be ok.

IgnacioBecerra (approved):
Ignacio approved ™️

alisonjoseph (approved):

PR #663, opened 2025-06-18
feat: add-tile-click-prop

seVishnu:
Allow user to pass a method that will be called while tiles are clicked.

#### Changelog

**New**

- tileClickHandler prop added


seVishnu:

seVishnu:
recheck

seVishnu:
> Please also add this prop to AnimatedHeader.mdx file where we show example with all props specified.



> Please also add this prop to AnimatedHeader.mdx file where we show example with all props specified.

this is done.

Issue #682, opened 2025-06-25
[Bug]: Cannot be used with other Carbon SCSS with import


lee-chase:
### Browser

Chrome

### Package version

0.47.0

### React version

19

### Description

Carbon SCSS imports into project clash with UI Shell import. This may be because the import order causes Component SASS (which can include layout, breakpoint, theme etc) to be processed before the following Carbon import which is the only SCSS in the identified file.

NOTE if I drill down into ui-shell.scss and comment out `styles/side-nav` the issue goes away.

```
@use "@carbon/react" with (
  $use-per-family-plex: true
);
@use "@carbon-labs/react-ui-shell/scss/ui-shell";
@use "@carbon/ibm-products/css/index-without-carbon-released-only.min";
```

![Image](https://github.com/user-attachments/assets/c531f099-04df-4ad3-bd08-61f7980b1e07)

### Reproduction/example

N/A

### Steps to reproduce

Not yet achieved preproduction outside of product.

### Suggested Severity

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-labs/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/carbon-labs/issues) for duplicate problems

lee-chase:
@alisonjoseph not been able to re-create this in stackblitz.

PR #683, opened 2025-06-25
fix(uishell): remove duplicate import

alisonjoseph:
Possible fix for #682

#### Changelog

**Removed**

- Removed duplicate breakpoint import from side-nav.scss


lee-chase:
Possible, would have to give it a try.

lee-chase:
So does not fix the issue for me but if move the UIShell SASS import to a script file, instead of with the Carbon SASS, it starts to work. My guess is that this is resulting in two separate compiles of the SCSS and hence no clash.

PR #691, opened 2025-06-30
fix(uishell): update style import for ibm-products

alisonjoseph:
Fix for odd style bug that showed up after we imported ibm-products scss
Screenshot 2025-06-30 at 11 39 43 AM

#### Changelog

**Changed**

- import only UserAvatar styles from ibm-products

#### Testing / Reviewing
Check Demo story


ariellalgilmore (approved):
AMAZING!!!

Issue #681, opened 2025-06-25
Styles loading out of order on first load in Storybook (resolved on refresh)


alisonjoseph:
There is an issue showing up across several components in Storybook where, due to the order in which style sheets are loading, certain components render incorrectly on first load. Refreshing the page resolves the issue. This only affects Storybook, not an issue with the actual components.

### Examples:

Theme switcher loads blue on first load
https://labs.carbondesignsystem.com/?path=/story/react_components-themesettings--default

<img width="188" alt="Image" src="https://github.com/user-attachments/assets/f894a82b-62d6-4ebe-b68b-c9018b84c16e" />

Button alignment is off on first load
https://labs.carbondesignsystem.com/?path=/story/react_components-uishell--default
<img width="313" alt="Image" src="https://github.com/user-attachments/assets/f7203c4a-9124-4439-b9f6-84b2726b4ccc" />

PR #697, opened 2025-06-30
fix(styles): remove unnessary button imports

ariellalgilmore:
Ref #681

Fixes part of the issue above so the Theme switcher doesn't load blue on first load

#### Changelog

**Removed**

-  remove unnessary button style imports being added

#### Testing / Reviewing

make sure other stories aren't effected, but i removed the import from resizer and split-panel... neither which use a button


alisonjoseph (approved):
🙌

PR #715, opened 2025-07-18
feat(animated-header): add accessibility role and aria- attributes

### LINKED ISSUE #714: [a11y]: Review accessibility role and aria- attributes for Animated Header
### Browser

_No response_

### Operating System

_No response_

### Package version

0.22.0

### React version

_No response_

### Automated testing tool and ruleset

wdp-gov team

### Assistive technology

_No response_

### Description

Review accessibility role and aria- attributes for Animated Header

### WCAG 2.1 Violation

_No response_

### Reproduction/example

View in browser

### Steps to reproduce

In browser missing role and aria- attributes

### Code of Conduct

glapadre:
Closes #714 

This PR adds neccessary role and aria- attributes for animated header

#### Changelog

**New**

- N/A

**Changed**

- AnimatedHeader, GlassTile, AIPromptTile, TaskController, WorkspaceSelector

**Removed**

- N/A

#### Testing / Reviewing

Testing in browser


packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
```diff
@@ -29,6 +29,27 @@ import WorkspaceSelector, {
...
+  'zh', // Chinese
+];
+
+export interface AriaLabels {
+  welcome?: string;
+  description?: string;
+  collapseButton?: string;
+  expandButton?: string;
+  tilesContainer?: string;
+  tiles?: { [tileId: string]: string };
```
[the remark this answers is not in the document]
IronZDev: Won't it be an anti-pattern to set the same aria-labels for all tiles? Each tile should have a different one I believe.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
```diff
@@ -29,6 +29,27 @@ import WorkspaceSelector, {
...
+  tasksDropdown?: string;
+  workspaceDropdown?: string;
```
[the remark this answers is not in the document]
IronZDev: I would go with the first option and leave 
```
welcome?: string;
description?: string;
collapseButton?: string;
expandButton?: string;
tilesContainer?: string;

File: packages/react/src/components/AnimatedHeader/components/HeaderTitle/HeaderTitle.tsx [hunk 3]
```diff
@@ -0,0 +1,72 @@
...
+  ariaLabels,
+}: HeaderTitleProps) => {
+  const prefix = usePrefix();
+  const blockClass = `${prefix}--animated-header__title`;
+
+  const currentLang =
+    typeof window !== 'undefined'
+      ? document.documentElement.lang || 'en'
+      : 'en';
+  const isNameFirst = NAME_FIRST_LANGS.includes(currentLang.slice(0, 2));
```
jubitjoy, line 44: We can remove this once the issue(https://github.com/carbon-design-system/carbon-labs/issues/719) of multilingual support is resolved.

packages/react/src/components/AnimatedHeader/components/Tiles/GlassTile/GlassTile.tsx
```diff
@@ -69,6 +74,16 @@ export const GlassTile: React.FC<GlassTileProps> = ({
         glassTileClickHandler?.();
       }}
       className={`${prefix}--animated-header__tile ${blockClass}`}
+      aria-label={
+        ariaLabel ??
+        (isDisabled || isLoading
+          ? `${title} (disabled)`
+          : title
+          ? `Open ${title}`
```
jubitjoy, line 82: can we forward the title here ? without any transform.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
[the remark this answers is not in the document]
glapadre: Correct each Tile will always have it's own aria-label, checkout the guidance I added in the [Storybook](https://deploy-preview-715--carbon-labs-react.netlify.app/?path=/docs/components-animated-header--overview#accessibility). The TIle Container has a aria-label and each Tile has it's own aria-label.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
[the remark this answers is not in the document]
glapadre: @jubitjoy I agree — we shouldn’t transform any of the aria-labels, and I’ll remove the places where that’s currently happening.

As for using a string prop as an aria-label, we could treat it as a fallback, but not the primary source. It won't always be sufficient. For example, with a Carbon Button, just using label="Download" isn’t as descriptive as something like:

`<Button label="Download" aria-label="Download the report" />
`

So we will prioritize explicit aria-label values when provided.

packages/react/src/components/AnimatedHeader/components/Tiles/GlassTile/GlassTile.tsx
glapadre: Yes we can do that, but this is just for a default fallback. The first option is always going to be the provided `ariaLabel`.

packages/react/src/components/AnimatedHeader/components/AnimatedHeader/AnimatedHeader.tsx
```diff
@@ -26,9 +26,22 @@ import TasksController, {
...
+  [key: string]: any;
```
IronZDev, line 42: Why do we need that? I believe we won't use anywhere the arbitrary key values anyway

glapadre: There is no need I will remove.

IronZDev (approved):

PR #726, opened 2025-07-25
feat(SideNav): add pinning functionality

### LINKED ISSUE #652: [Platforms]: Side nav menu icon and pinning
Design issue https://github.com/carbon-design-system/carbon-labs/issues/630

Users need to be able to keep the local navigation persistently open through pinning. 
- From a composability perspective the pin is OPTIONAL
- From a Platform perspective we'd likely require or strongly suggest the pin functionality
- The pin will horizontally align to the decorative icon
- The pin is interactive, where the decorative icon is not
- See possible combinations example below

**Spec**
Image

**Possible combinations**
Image

**UX flow**
- Figma prototype: [Icon only](https://www.figma.com/proto/qBKGHpCZONxtUuYrlYVuMw/Platform-%E2%80%93-UI-shell?page-id=320%3A6625&node-id=321-16589&viewport=186%2C93%2C0.59&t=wv8Rlun1MOPnDlil-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=321%3A16589) (slightly buggy)

Image

emyarod:
Closes #652

This PR implements sidenav pinning functionality by enabling the `enableRailPin` prop. Options for custom icon rendering and click handlers can be configured fia the `sideNavPinProps` prop which accepts an object

#### Changelog

**New**

- `SideNavPin` component
- `enableRailPin` and `sideNavPinProps` props on `SideNav`

#### Testing / Reviewing

confirm that the side nav can be pinned and unpinned as expected according to the component spec

### DIFF
--- a/packages/react/src/components/UIShell/__stories__/UIShell.stories.js
+++ b/packages/react/src/components/UIShell/__stories__/UIShell.stories.js
@@ -70,7 +70,6 @@ import {
   Share,
   User,
   VirtualColumnKey,
-  Menu,
   IbmCloudKeyProtect,
   Group,
   Money,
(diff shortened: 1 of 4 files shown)


alisonjoseph:
🥲

PR #742, opened 2025-08-05
feat(SideNav): implement pin and auto expand

### LINKED ISSUE #730: [Platforms] User preference controls for rail hover and pinning
This enhancement provides users with the flexibility to control how they experience the local menu.

**Two user preference controls:**

1. Pin: Expands the local menu to be open, behaving as a grid influencer to push the page content over
2. Rail hover: Toggles between the fixed rail with tooltips and the rail that expands into a panel on hover

_See example for interaction_

https://github.com/user-attachments/assets/38fdbab3-4433-433d-b96d-fb116ed2ac80

---

- The selected state(s) need to be persistent for each user as they navigate through the platform and from session to session.
- Each action needs to be instrumented so that we can evaluate analytics to determine which local menu state is most commonly preferred by users (fixed rail vs expanding rail vs pinned menu).
- Designs leverage the same menu item components but use `label-01` type token instead of the default.
- If the menu is pinned open, the bottom 'rail control' item should be set to the disabled state, only to become active again once the user unpins the menu.

emyarod:
Closes #730

This PR adds side nav pinning and auto expand functionality

#### Changelog

**New**

- side nav pin toggle
- side nav auto expand toggle

**Changed**

- resolve useEffect dependency array lint errors
- update SideNavToggle to match SideNavLink

#### Testing / Reviewing

confirm that the new pinning and auto expansion features work as intended

### DIFF
--- a/packages/react/src/components/UIShell/__stories__/UIShell.stories.js
+++ b/packages/react/src/components/UIShell/__stories__/UIShell.stories.js
@@ -500,7 +500,7 @@ export const Demo = () => {
                     renderIcon={Layers}>
                     Infrastructure
                   </SideNavLink>
-                  <SideNavDivider />
+                  <SideNavTitle>Experience services</SideNavTitle>
                   <SideNavLink
                     href="http://www.carbondesignsystem.com"
                     renderIcon={Dashboard}>


alisonjoseph:
Couple things I'm seeing, but will want @cameroncalder to chime in here.

### Questions
- When the auto-expand is disabled, should the little triangle caret still appear in the rail view?
- When the nav is pinned should the disable link be disabled?
- We have this turned on as navType panel, does this naming still make sense?


### Todo
- Let's remove the `Demo` story and rename `Demo - Rail with flyouts` story to `Demo` since this will be the recommended approach. 
- Check the imports on the overview page and make sure nothing is missing
- Add navType panel docs to overview page for SideNav, make sure to add a note that it only works with 2 levels of nesting because of the flyout menus. Otherwise they have to use the rail variant only.

cameroncalder (commented):
- [ ] When the auto-expand is disabled, the triangle caret does not need to be displayed.
- [ ] When the nav is pinned open, the 'auto-expand' option should be disabled
- [ ] Need to add a divider above the bottom user preference controls
- [ ] When the menu is long enough to warrant a scrollbar, the bottom menu items are unable to be reached when scrolling.
- [ ] Styling update: Change the type token on the user preference controls (Pin / Auto-expand) to `$label-01` to hierarchically separate from the main menu items.
- [ ] The user preference controls appear to be behaving differently from the menu items when the rail expands to the hover. The icons shouldn't move and the text should reveal 

https://github.com/user-attachments/assets/95a110a7-a604-4eb7-ba5a-2a0f67cbc232

packages/react/src/components/UIShell/__stories__/UIShell.stories.js
```diff
@@ -1153,7 +638,7 @@ export const DemoPanel = () => {
                     renderIcon={Layers}>
                     Infrastructure
                   </SideNavLink>
-                  <SideNavTitle>Experience services</SideNavTitle>
+                  <SideNavDivider />
```
ariellalgilmore, line 641: Can we bring back the `SideNavTitle` guessing it got changed from the merge?

ariellalgilmore (commented):
looking good! just noticing the sidenavtitle disappeared

alisonjoseph:
Looks like a small linting issue with the order of some css styles

packages/react/src/components/UIShell/__stories__/UIShell.stories.js
emyarod: fixed! sorry about that

ariellalgilmore (approved):
LGTM! nice work!

cameroncalder (approved):
LGTM! 🚀

PR #751, opened 2025-08-13
feat: style-picker react wrapper

davidmenendez:
Closes https://github.com/carbon-design-system/ibm-products/issues/7709

Adds `react/` inside `style-picker/` for lit react wrapper components. Also includes an updated snapshot test.

#### Changelog

**New**

- `packages/web-components/src/components/style-picker/react/style-picker-color.ts`
- `packages/web-components/src/components/style-picker/react/style-picker-group.ts`
- `packages/web-components/src/components/style-picker/react/style-picker-icon.ts`
- `packages/web-components/src/components/style-picker/react/style-picker-option.ts`
- `packages/web-components/src/components/style-picker/react/style-picker-section.ts`
- `packages/web-components/src/components/style-picker/react/style-picker-sections.ts`
- `packages/web-components/src/components/style-picker/react/style-picker.ts`

**Changed**

- `packages/web-components/src/components/style-picker/__tests__/__snapshots__/style-picker.test.snap.js`

#### Testing / Reviewing

`yarn test`
`yarn test:updateSnapshot`

PR #770, opened 2025-08-26
feat: react StylePicker

davidmenendez:
Closes https://github.com/carbon-design-system/ibm-products/issues/7709

second attempt. original for reference: https://github.com/carbon-design-system/carbon-labs/pull/751

### DIFF
--- a/examples/react/StylePicker/index.html
+++ b/examples/react/StylePicker/index.html
@@ -1,0 +1,22 @@
+<!--
+@license
+
+Copyright IBM Corp. 2025
+
+This source code is licensed under the Apache-2.0 license found in the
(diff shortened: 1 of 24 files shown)


matthewgallo:
Something seems to be a bit off in the [deploy preview](https://deploy-preview-770--carbon-labs-react.netlify.app/?path=/story/components-stylepicker--default). Maybe the styles also need to be imported from the react style picker story file?

davidmenendez:
@matthewgallo if you look now it should resemble the wc version much more closely now https://deploy-preview-770--carbon-labs-react.netlify.app/?path=/story/components-stylepicker--default

PR #789, opened 2025-09-05
chore: update storybook publish workflow node version

alisonjoseph:
Has been failing for past two weeks, since storybook 9 upgrade https://github.com/carbon-design-system/carbon-labs/actions/workflows/storybook-publish.yml


tay1orjones (approved):
Should we add this to the slack notifications workflow so we're more likely to notice if it's failed?

alisonjoseph:
@tay1orjones probably wouldn't hurt

PR #806, opened 2025-09-16
fix: stylepicker replace icons

davidmenendez:
Closes https://github.com/carbon-design-system/ibm-products/issues/8251

updates icons in the `style-picker` web component from `@carbon/web-components` to `@carbon/icons`


matthewgallo (approved):
There appears to be a formatting issue in `_placeholders.ts` but otherwise looks great 🎉

PR #833, opened 2025-10-07
chore(animated-header): fix ts bugs

IronZDev:
This PR is to fix TS bugs and inconsistencies.

#### Changelog

**Changed**

- Fixed TS issues
- Fixed conditional hooks triggering

### DIFF
--- a/packages/react/src/components/AnimatedHeader/components/Tiles/BaseTile/BaseTile.tsx
+++ b/packages/react/src/components/AnimatedHeader/components/Tiles/BaseTile/BaseTile.tsx
@@ -17,13 +17,10 @@ import { GlassTile, type GlassTileProps } from '../GlassTile/GlassTile';
 /** Base Tile router */
 export type TileVariant = 'glass' | 'aiPrompt' | 'ai';
 
-export type BaseTileProps =
-  | ({ variant?: 'glass' } & GlassTileProps)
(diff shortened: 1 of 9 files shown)


packages/react/src/components/AnimatedHeader/components/Tiles/BaseTile/BaseTile.tsx
```diff
@@ -17,19 +17,13 @@ import { GlassTile, type GlassTileProps } from '../GlassTile/GlassTile';
...
-      tileId?: string | null;
-    });
+export type BaseTileProps = GlassTileProps | AIPromptTileProps | AITileProps;
 
 function inferVariant(props: BaseTileProps): TileVariant {
   if (props.variant) {
     return props.variant;
   }
-  if (props.id === 'ai-tile') {
+  if (props.tileId === 'ai-tile') {
```
glapadre, line 26: This is designed as a fallback before the variant prop was introduced. Before variant prop was introduced every Tile was a Glass Tile unless the Tile `id` was set to 'ai-tile' which wasn't ideal.. so the variant prop was added. So this needs to stay as just `id` and not `tildId` as a fallback for users who aren't using `variant` yet

glapadre (changes requested):

Issue #827, opened 2025-09-30
[Platforms]: UIShell html structure bugs


alisonjoseph:
### Summary

There are a couple of issues within **SideNav** when using `type="RAIL_PANEL"`.

---

### 1. Incorrect rendered HTML

The current rendered HTML structure for the SideNav is not correct. Specifically, **popovers and tooltips are not being rendered inside their `<li>` elements**.

**Expected HTML structure:**

```html
<ul>
  <li>...</li>
  <li>...</li>
  <li>...</li>
  ...
</ul>
```

---

### 2. Inline styles in `SideNavFlyoutMenu.tsx`

Inline styles are currently being used to control visibility in `SideNavFlyoutMenu.tsx`. These should be removed and replaced with class names and CSS.

**Current code:**

```tsx
<div style={{ display: isButtonFocused ? 'none
```

PR #847, opened 2025-10-14
fix(ui-shell): resolve html structure bugs

emyarod:
Related #827

This PR updates the sidenav popovers to render inside `li` elements as well as removes inline stules in `SideNavFlyoutMenu`

note: the referenced issue will require an upstream update for iconbuttons/tooltips

#### Changelog

**New**

- side nav flyout menu classes for display styles

**Changed**

- rendered sidenav popover and tooltip html

**Removed**

- inline styles

#### Testing / Reviewing

confirm that the sidenav popovers and tooltips are rendered properly and check that the styles for the flyout menus are unchanged


alisonjoseph:
@emyarod I'm seeing this html now, which is still invalid, wasn't sure exactly what part of this needed an upstream change?

```
<li class=popover...>
   <li class=side-nav__item....>
```

Screenshot 2025-10-15 at 8 55 49 AM

emyarod:
@alisonjoseph I thought that IconButton would need the `as` prop to pass into the underlying Tooltip for some of the sidenav items. I'll also look into the doubled `li > li`, but it should be fine if we're able to change the innermost `li` to a `span` or something right?

PR #854, opened 2025-10-16
feat(header-overflow-panel): add to ui-shell

### LINKED ISSUE #777: [Platforms] Implement S & MD breakpoints for UI Shell header components
The current implementation of the UI shell header for Small screen sizes needs to be updated based on the decisions from: https://github.com/carbon-design-system/carbon-labs/issues/778

### ExpandableSearch
- [x] Medium and Small breakpoints: Expand across entire Header

### New HeaderOverflowPanel (naming tbd) component
Create a new component that appears when clicking the overflow icon (three dots) in the header. This component should support nested navigation with a back button pattern, and can display rich content beyond simple menu items in the detail view.

```
HeaderOverflowPanel 
├── HeaderOverflowPanel.Menu (primary view with links)
├── HeaderOverflowPanel.Item (individual menu items)
├── HeaderOverflowPanel.Detail (detail/nested view)
└── HeaderOverflowPanel.Back (back button)
```

At Medium breakpoint 
- Everything except Search and Profile collapse into Overflow 

At Small breakpoint...
- TrialCountdown component is hidden 
- Everything except Search collapses into Overflow

## Questions
- Is the TrialCountdown hidden on mobile screens?

# Design specs
*Some of the images here in this issue are not rendering right, so refer to the figma file.

Ref: [Figma file](https://www.figma.com/design/CNDTu3Gghv5CuBF2L0Krf9/UI-Shell-Header-%E2%80%94-SM-and-MD-breakpoints?node-id=12-6597&t=XCQGNDjBTXvU0a5C-1)

---

## Small breakpoint

Image

Image

-----

## Medium breakpoint

Image

Image

Image

ariellalgilmore:
Closes #777 

Create new HeaderOverflowPanel. Created a HeaderOverflowPanel story and mdx files. Also added it to the demo story
#### Changelog

**New**

- HeaderOverflowPanel stories, styles, and mdx files
- updated Demo to use HeaderOverflowPanel

**Changed**

- changed classname from primary to secondary for the secondary items in the wrapper

#### Testing / Reviewing

 Check demo and HeaderOverflowPanel stories

### DIFF
--- a/packages/react/src/components/UIShell/components/HeaderOverflowPanel.tsx
+++ b/packages/react/src/components/UIShell/components/HeaderOverflowPanel.tsx
@@ -18,6 +18,9 @@ import {
   HeaderPopoverContent,
 } from '../components/HeaderPopover';
 import { OverflowMenuVertical } from '@carbon/icons-react';
+import { useMatchMedia } from '../internal/useMatchMedia';
+import { breakpoints } from '@carbon/layout';
+const mdMediaQuery = `(max-width: ${breakpoints.md.width})`;
 
 export interface HeaderOverflowPanelProps {
   /**
@@ -53,9 +56,10 @@ export const HeaderOverflowPanel = React.forwardRef<
     [`${prefix}--header-overflow-panel`]: true,
     [customClassName as string]: !!customClassName,
   });
+  const isMd = useMatchMedia(mdMediaQuery);
   return (
     <HeaderPopover ref={ref} align="bottom-end" className={className} {...rest}>
-      <HeaderPopoverButton align="bottom-end" label={label}>
+      <HeaderPopoverButton align={isMd ? 'bottom-end' : 'bottom'} label={label}>
         <OverflowMenuVertical />
       </HeaderPopoverButton>
       <HeaderPopoverContent>
--- a/packages/react/src/components/UIShell/components/Profile.tsx
+++ b/packages/react/src/components/UIShell/components/Profile.tsx
@@ -59,7 +59,7 @@ const Profile = React.forwardRef<HTMLDivElement, ProfileProps>(function Profile(
   });
   return (
     <HeaderPopover ref={ref} align="bottom-end" className={className} {...rest}>
-      <HeaderPopoverButton align="bottom" label={label}>
+      <HeaderPopoverButton align="bottom-end" label={label}>
         {IconElement}
       </HeaderPopoverButton>
       <HeaderPopoverContent>{children}</HeaderPopoverContent>
--- a/packages/react/src/components/UIShell/components/styles/_header-overflow-panel.scss
+++ b/packages/react/src/components/UIShell/components/styles/_header-overflow-panel.scss
@@ -64,6 +64,13 @@ $prefix: 'cds' !default;
       fill: $link-primary-hover;
     }
   }
+
+  .#{$prefix}--header-overflow-panel .#{$prefix}--side-nav__item {
+    .#{$prefix}--side-nav__link,
+    .#{$prefix}--side-nav__submenu {
+      block-size: $spacing-08;
+    }
+  }
 }
 
 @include breakpoint(md) {


laurenmrice (changes requested):
Hey @ariellalgilmore, this is looking good! Here are my changes/comments below:

### Medium breakpoint

**Overflow**
- On hover/focus, the tooltip should be the normal high contrast color (`$background-inverse`), unless there is a reason why it is the color it is now. Note: I may not have all the previous shell knowledge and design decisions Olivia/Aubrey made so let me know.
- For the tooltip, I would rename the text to say “Options”.

tooltip)


- You can remove the “Log out” menu item and the horizontal rule above it in the Medium breakpoint, because Log out is already in the Profile menu, which is present in this breakpoint.
- When the overflow menu is open, and you move focus to the first item in the menu, the overflow menu icon button needs to receive a border on the right and left sides using color token `$border-subtle`, so it creates a visual “notch” into the header. (I noticed this also needs to be fixed for the profile menu when it’s open, but that may be a separate issue.)

focus on item in menu)



- When the overflow menu is open, there needs to be a vertical border on the full left side of the menu using color token `$border-subtle`. This menu should extend flush to the bottom of the page, because we can not determine how much content will end up being in the menu, and if you can navigate through the menu to different menu pages, it could possibly create a “jumping” experience  of different menu heights depending how much content there is, which could be visually jarring. A consistent height for this menu could resolve that problem.

- The top of the menu when it is open is slightly obscuring the horizontal rule of the UI shell header. (This is also happening with the Profile menu).

Screenshot 2025-10-20 at 4 06 20 PM

- Use the “chart custom” icon for the “Custom action” menu item.
- When you click on the “Settings” item in the menu, it navigates to the next page of the menu and a “Back” button appears at the top of it. The interaction between these menu pages visually jumps a bit because the text of “Settings” and “Back” are not in the same place. They should both have the same amount of spacing at the top in each menu page.

Multiple menu pages type alignment

The specs for the second page of the menu with the back button should be like this:

Specs for second page

----

### Small breakpoint
Still need to review this! TBD

ariellalgilmore:
@laurenmrice updated the following

### Medium breakpoint

**Overflow**

- [ ] - On hover/focus, the tooltip should be the normal high contrast color (`$background-inverse`), unless there is a reason why it is the color it is now. Note: I may not have all the previous shell knowledge and design decisions Olivia/Aubrey made so let me know. 
     - this will be fixed once this get merged in https://github.com/carbon-design-system/carbon/pull/20669 
- [x] - For the tooltip, I would rename the text to say “Options”.
- [ ] - The Search icon button in the header doesn't receive a tooltip on hover/focus, but it probably should for a11y reasons like the other icon buttons in the header? 
     - will open a seperate issue in core, so this will be fixed upstrem eventually

- [x] - You can remove the “Log out” menu item and the horizontal rule above it in the Medium breakpoint, because Log out is already in the Profile menu, which is present in this breakpoint.
- [ ] -  When the overflow menu is open, and you move focus to the first item in the menu, the overflow menu icon button needs to receive a border on the right and left sides using color token `$border-subtle`, so it creates a visual “notch” into the header. (I noticed this also needs to be fixed for the profile menu when it’s open, but that may be a separate issue.)
   - Will open a separate issue for this


- [ ] -When the overflow menu is open, there needs to be a vertical border on the full left side of the menu using color token `$border-subtle`. This menu should extend flush to the bottom of the page, because we can not determine how much content will end up being in the menu, and if you can navigate through the menu to different menu pages, it could possibly create a “jumping” experience  of different menu heights depending how much content there is, which could be visually jarring. A consistent height for this menu could resolve that problem.
   - The border will also get fixed once this is merged in https://github.com/carbon-design-system/carbon/pull/20669. The height is currently always the full height of the screen. 

- [ ] -The top of the menu when it is open is slightly obscuring the horizontal rule of the UI shell header. (This is also happening with the Profile menu).
  - I think this can go in the same issue as above with the notch
- [x] - Use the “chart custom” icon for the “Custom action” and "Customize" menu items.
- [x] - When you click on the “Settings” item in the menu, it navigates to the next page of the menu and a “Back” button appears at the top of it. The interaction between these menu pages visually jumps a bit because the text of “Settings” and “Back” are not in the same place. They should both have the same amount of spacing at the top in each menu page.

----

ariellalgilmore:
@laurenmrice updated the sizing to 40px for md/sm breakpoints. The tooltip alignments are also updated, but search will be updated once the issue in core is contributed

PR #898, opened 2025-10-30
chore(deps): update to carbon react .95.0-rc.0

alisonjoseph:
#### Changelog
- update to latest @carbon/react and @carbon/styles


tay1orjones (approved):
In addition to correcting the version range for the styles package I also added a yarn dedupe to ci.

PR #915, opened 2025-11-11
fix(uishell): various console errors

alisonjoseph:
Fixed various console errors showing up in the stories for UI Shell.

#### Changelog

**Changed**

- SideNavLink can't be a child of SideNavMenu, updated to SideNavMenuItem
- Removed isFixed prop that doesn't exists for HeaderMenuButton
- Only pass highContrast to IconButton, not to custom components in HeaderPopover
- Make sure SideNavItems always wraps children for correct html

#### Testing / Reviewing

Make sure UI Shell still renders as expected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,9 @@ yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
 .yarn/install-state.gz
-
+launch.json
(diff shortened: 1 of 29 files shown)


emyarod (approved):
looks good to me! just one question about the header popover

PR #1186, opened 2026-04-16
feat(TagInput): New react compoenet - initial commit 

punnoosewilson:
# TagInput
The TagInput component provides an intuitive interface for users to add, manage, and remove tags. It combines a text input field with tag visualization, enabling users to create collections of labeled items through keyboard interaction. Tags are displayed inline with the input field, and users can navigate between tags using arrow keys, making it accessible and efficient for data entry.

#### Changelog

**New**

- Included a new react component (TagInput)

related
https://github.com/carbon-design-system/carbon/issues/21570
https://github.com/carbon-design-system/carbon/issues/10047


devadula-nandan (approved):
I have 1 optional suggestion which can be tackled later too depending on a11y/design of this. and are good to have's

https://github.com/user-attachments/assets/bc9026af-7a9d-4a7f-9136-ee43dd8ec00c

While I can focus the tags using the arrow keys while I'm in the input field, I currently have to press Delete to remove them. Is there a way to keep the focus only on the dismissible button and avoid making the tags themselves tabbable, so it matches the existing dismissible tag component?

We could still support arrow key navigation by moving focus directly to the dismissible button, allowing it to be triggered with Space, Enter, or Backspace. Is this possible, or would it introduce any challenges from A11Y pov?

I really like the [With Suggestions](https://deploy-preview-1186--carbon-labs-react.netlify.app/?path=/story/components-taginput-examples--with-suggestions) use case. i remember someone asking this in our office hours.

Issue #1181, opened 2026-04-16
[UI Shell Header] Bug fixes


Kritvi-bhatia17:
## Description
After auditing the current UI Shell Labs implementation against the Figma specifications, we identified a few inconsistencies across the header and specific components, outlined below.

### Unexpected hover stroke
An extra border/stroke appears when hovering over UI shell elements (action items, menus, profile, etc.).
Per design specs, elements should not highlight their borders on hover.
- [ ] This likely needs investigation at the code level to identify the root cause and then to be removed.
    > <img width="588" height="230" alt="Image" src="https://github.com/user-attachments/assets/93bc474e-a775-44ea-961a-946a681f3652" />

### Incorrect border token
- [ ] Update all the UI shell header borders and dividers color token as $border-subtle(contextual) / $border-subtle-01. Currently, this appears to be set to $border-subtle-00.

### Divider
- [ ] Update the divider's width in Labs Storybook to be 1px wide. [Figma specifications](https://www.figma.com/design/EsmsdPe4MO8iJeyQV78K25/UI-Shell-Enhancements?node-id=2018-12634&t=5uFwL8UYF1xeFYt9-11) explicitly call for a 1px divider.

### Trial Countdown component
- [ ] There is a white inner border on the [trial countdown component](https://labs.carbondesignsystem.com/?path=/story/react_components-uishell--demo) ( more evident in light theme) when focused + hovered. This doesn’t seem expected—can we identify the source and see if it’s being overridden and removable?
    > <img width="3456" height="2234" alt="Image" src="https://github.com/user-attachments/assets/542c2ff3-507f-4bc8-83dc-012171fd80b7" />

### Search
- [ ] Icon size mismatch: Update the search icon size to be 20px from 16px, to align with other header action icons. (Since the Search component enforces a max icon size of 16px, this might require a component customization specifically for UI shell header).
- [ ] Incorrect icon color: Update the search icon color to uses $icon-primary to match with other header utilities. Figma spec link

### Environment Switcher (Menu & Sub‑menu)
- [ ] Menu selected state: In the current implementation, the border is rendered outside the menu instead of inside.
    > <img width="352" height="324" alt="Image" src="https://github.com/user-attachments/assets/49f2f8b0-6508-4adb-82f7-79c4b3745c34" />

PR #1272, opened 2026-06-09
feat(ui-shell): ui-shell header enhancements

### LINKED ISSUE #1192: [UI Shell Header] Enhancements
## Description
After auditing the current UI Shell Labs implementation against the Figma specifications, we identified a few enhancements across the header and specific components, outlined below.

### Trial Countdown component
- [ ] Text color updates
     > Number text color should use `$text-primary`
     > Label text color should use `$text-secondary`
     > Image 

- [ ] Additional states: Introduce more states for the Trial Countdown component as defined in the [Figma specs](https://www.figma.com/design/jB8VhMcWAQODSBpOWVSNBf/UI-Shell-%E2%80%94-Composable-header-specs?node-id=481-27322&t=nqIhiPXcRVI97Boa-11).

### All menus and popover content
- [ ] Add a 1px border and shadow to all menu and popover surfaces in the header:
     > Border token: $border-subtle-01
     > Apply $menu shadow for consistency with other menu surfaces

### Environment switcher (menu & sub-menu)
- [ ] Menu background styling: The background color of the Environment Switcher menu and sub-menu (when open) should match other menu panels in the header for visual consistency:
     > Open menu background: $layer-background(contextual) / $layer-background-01 | [Figma specs](https://www.figma.com/design/jB8VhMcWAQODSBpOWVSNBf/UI-Shell-%E2%80%94-Composable-header-specs?node-id=495-27373&t=nqIhiPXcRVI97Boa-11)
- [ ] Open menu alignment: All the menus will open with End alignment.
     > Image
- [ ] Remove bottom divider: The bottom divider highlighted in the below image should be removed, as it is inconsistent with other open menus in the UI Shell header.
     > Image

### Profile
- [ ] The [profile component in labs](https://labs.carbondesignsystem.com/?path=/docs/react_components-uishell-profile--overview) should support different profile types ((like initials or user avatars)), similar to what we have in [C4IP](https://ibm-products.carbondesignsystem.com/?path=/story/components-useravatar--default)?

maradwan26:
Closes #1192

Adds various enhancements to the UI Shell Header.

#### Changelog

**New**

- **Trial Countdown**
    - Aligned tokens of label text and number text with design spec
    - Fixed some spacing and centering issues
    - Aligned multiple states with the design spec: (e.g. hover, active)
        - Fixed an issue with the trial-countdown__count not taking in the colour of its background when in any state that wasn't the default `:enabled` state. 

- **All menus and popover content**
    - Added `$border-subtle-01` border and a `box-shadow`

- **Environment switcher (menu & sub-menu)**
    - Applied an alignment of `bottom-end` to the MenuButtons in the story
    - Aligned colour tokens
    - Added a border around the menu and menu-button when in an open state

- **Profile**
    - Added new props `image`, `imageDescription`, and `renderIcon` to ProfileUserInfo
        - ProfileUserInfo uses [UserAvatar from C4IP](https://ibm-products.carbondesignsystem.com/?path=/story/components-useravatar--default) internally, so all that was done here was just allowing the user to pass these new props through to UserAvatar

#### Testing / Reviewing
- All enhancements done should have no visual issues and align with the design specs and completion requirements outlined in the attached issue


heloiselui (commented):
Looking good. I just have a question about this border, in Kritvi’s image it looks like this:

Image

Just want to double check, what we have below, is this what we expect?

Image

-----

We removed the bottom divider, but the corner appears a bit different:

Image

Kritvi-bhatia17 (changes requested):
Hi @maradwan26, thanks for working on it, great work!

Few comments I wanted to share, happy to discuss those over a call if you need.

**Trial countdown**
1. The Trial Countdown looks different in its hover state when the menu is open versus when it's closed. The same applies to the other menu items—ideally, the behavior should be consistent with the Environment Switcher pattern.
2. Which shadow token is currently being used for the Trial Countdown menu when it's open? It seems slightly different from the Deploy Preview link, although that could just be how it appears to me.
3. When the Trial Countdown menu is open, the focus border should appear above the gray border, matching the behavior shown in the [Figma specs](https://www.figma.com/design/jB8VhMcWAQODSBpOWVSNBf/UI-Shell-%E2%80%94-Composable-header-specs?node-id=483-26895&t=Syitovgf0HS3xUG1-11) (see image below).
   > Image

---

**All menus and popover content**

4. There seems to be a small gap in the border whenever a menu is open. Could you please double-check that?
   > Image
5. Also, what border color token is currently being used for the UI Shell bottom border? Ideally, it should match the border color used for the popover/menu for better visual consistency.

---

**Environment switcher (menu & sub-menu)**

6. The border in the Environment Switcher doesn't seem to align properly with the UI Shell header. I've attached an image to illustrate the issue—could you please take a look and check what's causing it? 

   > Image

---

**Profile**

   > ProfileUserInfo uses UserAvatar from C4IP internally, so all that was done here was just allowing the user to pass these new props through to UserAvatar
7. When you say it's using UserAvatar from C4IP, isn't that already what's defined within this [Lab's Profile component](https://labs.carbondesignsystem.com/?path=/docs/react_components-uishell-profile--overview)? Just trying to understand if there's a different implementation being referenced here.

---

**General**

8. Any idea why the tab order is skipping the Content Switcher on the left (the 9-dot menu) and the Trial Countdown component, and going directly to Search instead? Is this intentional behavior, or could it be a bug?

maradwan26:
@Kritvi-bhatia17 

> **Trial countdown**
> 
> 1. The Trial Countdown looks different in its hover state when the menu is open versus when it's closed. The same applies to the other menu items—ideally, the behavior should be consistent with the Environment Switcher pattern.

Fixed 👍 

> 2. Which shadow token is currently being used for the Trial Countdown menu when it's open? It seems slightly different from the Deploy Preview link, although that could just be how it appears to me.

There's no token currently used for that popover content, Trial Countdown is used inside of HeaderPopover and its shadow just comes from Popover's dropshadow class from the core library which uses a custom `$popover-drop-shadow` variable that returns `drop-shadow(0 $spacing-01 $spacing-01 rgba(0, 0, 0, 0.2))`

The shadow I use in this PR's deploy preview is the `box-shadow` mixin from the carbon/styles utilities which returns `box-shadow: 0 2px 6px theme.$shadow` which returns `0 2px 6px rgba(0, 0, 0, 0.8)`

So the opacity of the shadow is darker in the deploy preview, from 0.2 to 0.8.

If there's a different token I should be using please let me know.

> 3\. When the Trial Countdown menu is open, the focus border should appear above the gray border, matching the behavior shown in the [Figma specs](https://www.figma.com/design/jB8VhMcWAQODSBpOWVSNBf/UI-Shell-%E2%80%94-Composable-header-specs?node-id=483-26895&t=Syitovgf0HS3xUG1-11) (see image below).

Fixed 👍 

> **All menus and popover content**
> 
> 4. There seems to be a small gap in the border whenever a menu is open. Could you please double-check that?

Fixed 👍 

> 5. Also, what border color token is currently being used for the UI Shell bottom border? Ideally, it should match the border color used for the popover/menu for better visual consistency.

The colour token for the bottom border of the UIShell header is [actually set in the Core library](https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/ui-shell/header/_header.scss#L36), It uses `border-subtle` while the borders set in this PR use `border-subtle-01`. Should this be done in a follow-up PR in the Core repo?


> **Environment switcher (menu & sub-menu)**
> 
> 6. The border in the Environment Switcher doesn't seem to align properly with the UI Shell header. I've attached an image to illustrate the issue—could you please take a look and check what's causing it?

https://github.com/user-attachments/assets/563009bc-bfce-4ce6-8a87-15deeb7546e5

It looks aligned for me but can misalign sometimes if I mess with the zoom percentage. I've noticed odd behaviours like this with some other components I'm not sure what causes this to happen. Most notably being the Popover TabTip variant will get an underline if you zoom out a bit:

https://github.com/user-attachments/assets/d89f29a4-968f-4405-80b5-2d1113dc07ce

I'm not sure if someone has noticed these issues before or if it's been discussed if this is an issue with our styles or some browser behaviour


> **Profile**
> 
> 7. When you say it's using UserAvatar from C4IP, isn't that already what's defined within this [Lab's Profile component](https://labs.carbondesignsystem.com/?path=/docs/react_components-uishell-profile--overview)? Just trying to understand if there's a different implementation being referenced here.

Yeah so basically Profile uses UserAvatar from C4IP internally, so some of the inputted props for Profile get passed forward to UserAvatar. However, before this PR, Profile would not let you use a prop to pass in an image for the avatar, which UserAvatar supports.

So all that was done here was just adding a prop to allow for images to be used for the avatar inside of Profile, and just passing it to the internal UserAvatar.


> **General**
> 
> 8. Any idea why the tab order is skipping the Content Switcher on the left (the 9-dot menu) and the Trial Countdown component, and going directly to Search instead? Is this intentional behavior, or could it be a bug?

It goes in the correct order for me, I think most likely you could've been clicking the empty space between the TrialCountdown and the Search to focus the UIShell before tabbing? The blank space in between is the container of those actions so clicking it would make the Search the next item in tab order. This is the only way I was able to reproduce this so that's what I think might've happened.

Screenshot 2026-06-25 at 12 19 19 PM


https://github.com/user-attachments/assets/e31c7e51-4aa1-4e39-94d0-22b07a20551f

maradwan26:
@heloiselui 

> Just want to double check, what we have below, is this what we expect?

> image


I interpreted the design spec as an underline for the TrialCountdown, which would then appear above the UIShell Header's border line.

@Kritvi-bhatia17 Can you clarify if this is the correct interpretation? I can remove the underline otherwise.

> We removed the bottom divider, but the corner appears a bit different:

Yeah the menu has `end` alignment now, but 'Select Category' is a quite a long label so the corner only gets moved a tiny bit like in the image.

Kritvi-bhatia17:
> There's no token currently used for that popover content, Trial Countdown is used inside of HeaderPopover and its shadow just comes from Popover's dropshadow class from the core library which uses a custom $popover-drop-shadow variable that returns drop-shadow(0 $spacing-01 $spacing-01 rgba(0, 0, 0, 0.2))
> 
> The shadow I use in this PR's deploy preview is the box-shadow mixin from the carbon/styles utilities which returns box-shadow: 0 2px 6px theme.$shadow which returns 0 2px 6px rgba(0, 0, 0, 0.8)
> 
> So the opacity of the shadow is darker in the deploy preview, from 0.2 to 0.8.
> 
> If there's a different token I should be using please let me know.
2. Is there a reason why is it mixin from the carbon/styles utilities, because the [correct specs](https://carbondesignsystem.com/components/menu/style/#color) is the drop-shadow(0 $spacing-01 $spacing-01 rgba(0, 0, 0, 0.2)) which you have mentioned earlier?
image

---
> The colour token for the bottom border of the UIShell header is [actually set in the Core library](https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/ui-shell/header/_header.scss#L36), It uses border-subtle while the borders set in this PR use border-subtle-01. Should this be done in a follow-up PR in the Core repo?
5. Oh, I have already documented that observation in [this issue](https://github.com/carbon-design-system/carbon-labs/issues/1181#issue-4276349195), so yes it will be taken care in that PR, thanks!
---

> It looks aligned for me but can misalign sometimes if I mess with the zoom percentage. I've noticed odd behaviours like this with some other components I'm not sure what causes this to happen. Most notably being the Popover TabTip variant will get an underline if you zoom out a bit:
> I'm not sure if someone has noticed these issues before or if it's been discussed if this is an issue with our styles or some browser behaviour
6. I think, in that case maybe skip for now since all of us has a lot of work at the moment, If possible can you please document that somewhere.

heloiselui (approved):
LGTM! Thanks @maradwan26. ~@Kritvi-bhatia17 could you check this part for us https://github.com/carbon-design-system/carbon-labs/pull/1272#issuecomment-4801929873?~ Done ✅

Kritvi-bhatia17:
> @heloiselui
> 
> > Just want to double check, what we have below, is this what we expect?
> 
> I interpreted the design spec as an underline for the TrialCountdown, which would then appear above the UIShell Header's border line.
> 
> @Kritvi-bhatia17 Can you clarify if this is the correct interpretation? I can remove the underline otherwise.
> 
> > We removed the bottom divider, but the corner appears a bit different:
> 
> Yeah the menu has `end` alignment now, but 'Select Category' is a quite a long label so the corner only gets moved a tiny bit like in the image.

Good call @heloiselui! Let's remove the trial-countdown component border @maradwan26. In the Figma specs, I had added that border because it was overlapping with the UI Shell border, and I wanted to highlight that it would still be visible on hover. However, since the entire UI Shell already has a border and the hover state is still clearly conveyed, it's fine to remove the additional border.

maradwan26:
@Kritvi-bhatia17 

Hello! I've made the following changes:

- Trial countdown underline has been removed
- `box-shadow` tokens have been removed in favour of what is coming from the core library
- Added top border to popover contents
- Created an issue for component misalignments on specific zoom percentages in the core repo so we can look further into what's causing this: https://github.com/carbon-design-system/carbon/issues/22550

Kritvi-bhatia17 (approved):
LGTM, great work Mahmoud! 🚀

devadula-nandan (approved):

PR #1287, opened 2026-06-23
fix(ui-shell): make useMatchMedia SSR-safe to prevent hydration mismatch

### LINKED ISSUE #1286: [SideNav]: React hydration mismatch on `inert` during SSR
When the `SideNav` is server-rendered, React can  report a hydration mismatch on the `<nav />` `intert` attribute.

<details>
<summary>Mismatched server/client properties</summary>

```
  - A server/client branch `if (typeof window !== 'undefined')`.
  - Variable input such as `Date.now()` or `Math.random()` which changes each time it's called.
  - Date formatting in a user's locale which doesn't match the server.
  - External changing data without sending a snapshot of it along with the HTML.
  - Invalid HTML tag nesting.
  It can also happen if the client has a browser extension installed which messes with the HTML before React loaded.
  https://react.dev/link/hydration-mismatch
    ...
      <div className="cds--g10 c...">
        <div style={{display:"flex", ...}}>
          <SideNavProvider>
            <Theme theme="white">
              <div className="cds--white...">
                <ShellLayout sideNavData={{...}}>
                  <CdsShellHeader>
                  <CdsSideNavClient gettingStartedSections={[...]} coreComponents={[...]} productComponents={[...]} ...>
                    <div>
                    <div ref={{current:null}} className="cxp--side-...">
                      <SideNavRenderFunction isFixedNav={true} isTreeview={true} expanded={false} isChildOfHeader={false} ...>
                        <nav
                          role="navigation"
                          tabIndex={-1}
                          ref={function useMergedRefs.useCallback}
                          className="cds--side-nav__navigation cds--side-nav cds--side-nav--collapsed cds--side-nav--wit..."
    +                       inert={undefined}
    -                       inert=""
                          aria-label="Side navigation"
                          aria-labelledby={undefined}
                          onFocus={function}
                          onBlur={function}
                          onKeyDown={function}
                        >
          ...

            [browser] Encountered two children with the same key, `.$/building-blocks/core/patterns/overflow-content`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```
</details>

The `inert` attribute is derived from a media query result here:
https://github.com/carbon-design-system/carbon-labs/blob/f7a3ab129162ae27ed39a03f0bc47d0621fa542a/packages/react/src/components/UIShell/components/SideNav.tsx#L628

The hook reads the live match during first render. On desktop viewports, the server renders `isLg = false` (`inert=""`), while client hydration render reads `isLg = true` (`inert={undefined}`). This is the mismatch. This is a hook-level bug that travels downstream to every server-rendered component.

kennylam:
Closes #1286 

During first client renders, `useMatchMedia` reads `window.matchMedia().matches`, while returning `false` on the server. This creates a hydration mismatch between the client and the server, e.g. `inert=""` on the server vs `inert={undefined}` o nthe client at > `lg`.

This changes the hook to use `useSyncExternalStore`, React's recommended primitive for external and SSR states. Since this fix is at the hook level, it should resolve the issue for all downstream users.

#### Changelog

**New**
- `useMatchMedia-test.js`: SSR/hydration regression coverage; added `TextEncoder`/`TextDecoder` polyfill to `jest.setup.ts` so react-dom/server can run under jsdom

**Changed**

- update `useMatchMedia` to use `useSyncExternalStore` with a `false` snapshot

**Removed**

- remove `canUseDOM`'s `useState` initializer and the manual `useEffect`

#### Testing / Reviewing

Testing SSR pages isn't possible in Storybook. Unit tests have been added to mock the conditions. All tests should pass.


maradwan26 (approved):
LGTM

PR #1301, opened 2026-07-09
Focus region

tombrunet:
Closes #

Adds `FocusRegion` — a React component and matching web component that enable F6/Shift+F6 keyboard navigation between major page regions, following the ARIA authoring practices landmark navigation pattern.

#### Changelog

**New**

- `FocusRegion` React component (`@carbon-labs/react-focus-region`) — wraps any page section and enrolls it in F6/Shift+F6 region navigation
- `useFocusRegion` hook — registers a region with the global registry and tracks focus location memory within it
- `useFocusGroupModal` hook — manages a modal stack so F6 navigation can be restricted to a specific group of regions when a dialog is open; exposes `pushModal()` / `popModal()`
- `NavigationHint` React component — shows a one-time inline hint on first keyboard focus explaining F6 navigation; controlled via `showNavigationHint` prop and persisted in `sessionStorage`
- `FocusRegion` SCSS — styles for the region focus indicator and navigation hint overlay
- React Storybook stories: Default, WithModal, WithDisabledRegions, FocusLocationTracking, NestedRegions, WithNavigationHint
- `clabs-focus-region` web component (`@carbon-labs/wc-focus-region`) — light-DOM Lit implementation with the same behaviour as the React version
- Shared `focus-region-registry.ts` — framework-agnostic singleton containing the region registry, modal stack manager, and global F6 keyboard handler; used by both the React hooks and the web component
- Web component Storybook stories using Carbon web components (`cds-button`, `cds-tile`, `cds-text-input`, `cds-stack`, `cds-link`): Default, WithDisabledRegion, FocusLocationTracking, WithNavigationHint
- Navigation hint in the web component uses `cds-actionable-notification`
- Stackblitz example app under `examples/react/FocusRegion`
- Unit tests for both the React and web component implementations

**Changed**

- Nothing.

**Removed**

- Nothing.

#### Testing / Reviewing

**React — start the React Storybook:**
```bash
cd packages/react
yarn storybook
```
Navigate to **Components → FocusRegion** at http://localhost:6006

**Web component — start the WC Storybook:**
```bash
cd packages/web-components
yarn storybook
```
Navigate to **Components → FocusRegion** at http://localhost:6007

---

- [ ] **Default (both)** — Tab moves naturally through all elements in DOM order with no trapping. Press **F6** to cycle forward through regions (Header → Navigation → Main → Footer → wraps). Press **Shift+F6** to cycle backwards.
- [ ] **FocusLocationTracking (both)** — Tab to a specific button or input in a region, press F6 to leave, press F6 again to return — focus lands on the last element visited in that region, not the first.
- [ ] **WithDisabledRegions (both)** — disable the sidebar; confirm it is skipped during F6 cycling. Re-enable; confirm it returns to the cycle.
- [ ] **WithNavigationHint (both)** — Tab (keyboard only, not mouse click) into any region for the first time; confirm the navigation hint notification appears. Dismiss via "Got it" or ×; confirm focus returns to the triggering element. Tab away and back — hint does **not** reappear. To reset: DevTools → Application → Session Storage → delete `clabs-focus-region-hint-shown` (React: `focus-region-hint-shown`).
- [ ] **WithModal (React only)** — open the modal; confirm F6 is restricted to the two modal regions. Close the modal; confirm F6 returns to the full page region cycle.
- [ ] **NestedRegions (React only)** — confirm F6 skips the nested complementary region and only cycles through the top-level banner, main, and contentinfo regions.


tombrunet:
There is a design question mark on this. The react version is automatically focusing the first element in the region. The web component version focuses the region first and then tabs from there. Needs some discussion regarding which version is more appropriate.

PR #1307, opened 2026-07-16
feat: add lazy loading for components which needs inital focus on page load

Gururajj77:
Closes https://github.com/carbon-design-system/carbon-experience-platform/issues/1818

Fix `StorybookDemo` iframe focus stealing and scroll jank caused by stories that programmatically focus elements on mount (e.g. `ActionableNotification`, `Popover` with `autoAlign`).

#### Changelog

**New**

- Added opt-in lazy loading to `StorybookDemo` — appending `--lazy` to a variant ID (e.g. `components-menu--default--lazy`) defers the iframe `src` until the iframe enters the viewport via `IntersectionObserver`. The suffix is stripped before the Storybook URL is built so the actual story ID is unaffected. The `lazy` boolean on the variant shape also works for programmatic use.
- Added `TabIndexTest` story — verifies that `ActionableNotification` embedded in an iframe below the fold does not steal host focus
- Added `TabIndexTestPopover` story — verifies that `Popover` (experimental auto-align) embedded in an iframe below the fold does not cause host page scroll jank
- Added `NoLazyLeakTest` story — verifies that plain Button variants (no `--lazy` suffix) load immediately and lazy loading does not leak onto variants that have not opted in

**Changed**

- Removed `allow-same-origin` from the `<iframe>` `sandbox` attribute — it had no effect since the host and iframe are always different origins, but its removal makes the intent explicit
- `StorybookDemo` variant shape now accepts an optional `lazy` boolean field

**Removed**

- Nothing.

#### Testing / Reviewing

- Open **"Tab index test (iframe below fold)"** story, scroll down, and Tab through the page — focus should skip all elements inside the `ActionableNotification` iframe
- Open **"Tab index test — popover (iframe below fold)"** story, scroll down — the host page should not scroll or jank when the Popover story mounts
- Open **"No lazy leak test — button (iframe below fold)"** story, open DevTools → Network, filter by `iframe.html` — the request should appear on page load before you scroll, confirming lazy loading does not apply to variants without the `--lazy` suffix
- To opt a new story into lazy loading, append `--lazy` to the Variant ID in the Payload CMS dashboard (e.g. `components-menu--default--lazy`)


elycheea (commented):
@Gururajj77 Did we still need `--lazy` suffix? I think we agreed to move `lazy` to a prop rather than variant based on the URL.

The fluid condensed fix looks fine although if this is the only prop, we can probably just set `isCondensed` inline for now?

Gururajj77:
> @Gururajj77 Did we still need `--lazy` suffix? I think we agreed to move `lazy` to a prop rather than variant based on the URL.
> 
> The fluid condensed fix looks fine although if this is the only prop, we can probably just set `isCondensed` inline for now?

Yep, made the changes! The PR https://github.com/carbon-design-system/carbon-experience-platform/pull/1899 takes care of adding the new `lazy` prop. 

And added `isCondensed` as inline prop.

packages/react/src/components/MDXComponents/components/storybook-demo/storybook-demo.tsx
```diff
@@ -168,7 +159,7 @@ export const StorybookDemo: MdxComponent<StorybookDemoProps> = ({
               onChange={onThemeChange}
               initialSelectedItem={themeItems[0]}
               className={border}
-              {...fluidProps}
+              isCondensed={!!fluid}
```
elycheea: Actually, now wondering if the live demo even needs a non-fluid and non-condensed version. 🙈 (Sorry, I didn’t think of this earlier. I think your last fix was right when I got back so I didn’t have full context yet!)

I think the Gatsby and new website always expect the fluid variant. So I wonder if we could _just_ set these to default to fluid instead? Especially since the non-fluid variant looks incorrect anyway.

Image

packages/react/src/components/MDXComponents/components/storybook-demo/storybook-demo.stories.jsx
```diff
@@ -94,11 +94,13 @@ WithFluidDropdown.args = {
 const actionableNotificationVariants = [
   {
     label: 'Default',
-    variant: 'components-notifications-actionable--default--lazy',
+    variant: 'components-notifications-actionable--default',
+    lazy: true,
```
elycheea: Instead of at the variant level, I think these should be set top level since it’s meant for the entire demo and not the specific story. Unless this is different than how you’ve approached it in Payload?

I see this as being a modifier at the same level as `wide` and `tall`. But open to your thoughts if this _should_ be based on a per-story implementation? It feels reasonable to me that if one variant requires lazy loading, we allow the entire demo to support it instead of constantly changing the demo behavior based on the story?

Gururajj77: Agreed, moved lazy to a top-level prop alongside wide and tall. The per-variant approach required looking up the current variant definition on every render to decide iframe behaviour.

packages/react/src/components/MDXComponents/components/storybook-demo/storybook-demo.tsx
Gururajj77: Yep, done

elycheea (approved):
Looks good to me. I _do_ think it’d be nice to hide these stories that are more for test scenarios though. (I think it was like a dev or test tag? but not sure that was something Core was doing.)

I don’t know if we have that set up in the Labs repo right now though.

heloiselui (approved):
LGTM. I think we can hide or remove these stories since they’re only for testing purposes. 

In Core we usually add test stories and then remove them afterwards.

Gururajj77:
> LGTM. I think we can hide or remove these stories since they’re only for testing purposes.
> 
> In Core we usually add test stories and then remove them afterwards.

Yep, have removed other stories, and have kept only default, since we made fluid as default for the `StorybookDemo`.

PR #1329, opened 2026-07-29
fix: move ref from iframe to link

Gururajj77:
Closes #1307 
> [!NOTE]
> remove story before merge

Move IntersectionObserver target from `<iframe>` to `<Link>` to prevent scroll jump

#### Changelog

**Changed**

- `StorybookDemo` lazy loading now observes the `full demo` link instead of the iframe — the iframe `src` is only set once the entire component has scrolled into view, preventing focus-stealing stories from jumping the page on load

**Removed**

- `iframeRef` removed from `<iframe>` — no longer needed

#### Testing / Reviewing

- Open the **Lazy load test (iframe below fold)** story in Storybook
- Scroll down — confirm the page does **not** jump to the component on load
- Check the Network tab before scrolling to confirm no iframe request fires until the `full demo` link enters the viewport

### DIFF
--- a/packages/react/src/components/MDXComponents/components/storybook-demo/storybook-demo.stories.jsx
+++ b/packages/react/src/components/MDXComponents/components/storybook-demo/storybook-demo.stories.jsx
@@ -81,3 +81,53 @@ Default.args = {
   tall: false,
   lazy: false,
 };
+
+const actionableNotificationVariants = [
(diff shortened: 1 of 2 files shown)


jlongshore (approved):
LGTM, tested in browser with the network tab opened and I observed the component load in.

PR #1334, opened 2026-08-04
chore(animated-header): add ssr support

IronZDev:
Adds proper SSR support to AnimatedHeader

#### Changelog

**New**

- Added `postbuild` script to `AnimatedHeader` package that writes a `lib/package.json` with `"type": "commonjs"` to fix SSR/CJS compatibility

**Changed**

- `AnimatedBackground`: replaced synchronous `window.matchMedia` call at render time with a `useState`/`useEffect` pattern, fixing SSR crashes and adding live reactivity to `prefers-reduced-motion` changes
- `AnimatedHeader`: removed the `typeof window !== 'undefined'` guard wrapping `<AnimatedBackground>` — SSR safety is now handled inside the component itself
- `HeaderTitle`: replaced synchronous `document.documentElement.lang` access with a `useState`/`useEffect` pattern, fixing SSR crashes

**Removed**

- Nothing.

#### Testing / Reviewing

- Verify `AnimatedHeader` renders without errors in an SSR environment (e.g. Next.js)
- Toggle OS `prefers-reduced-motion` setting while the header is visible and confirm the animation pauses/resumes without a page reload
- Confirm the header animation still plays correctly in a standard client-side render
- Confirm `HeaderTitle` renders correctly for both LTR and RTL/name-first languages (e.g. Japanese)


glapadre (approved):
LGTM

PR #1348, opened 2026-08-19
feat(react): implement Processing v2.0 proposal

mikeolasov:
Closes #

No changes

#### Changelog

**New**

Adds a brand new WAAPI-based Processing element with new functionality

**Changed**

Changes the component story to include the Processing v2.0 proposal

**Removed**

No changes

#### Testing / Reviewing

Check the interactive functionality, review package size and code quality/compliance.

PR #1375, opened 2026-09-18
feat(processing): Processing v2 — shape formations, declarative mode and promise-based handle

mikeolasov:
Supersedes #1348, which GitHub closed automatically after an unrelated branch was pushed over it by mistake.

Processing v2 moves the dots from CSS keyframes to the Web Animations API and adds shape formations (triangle, square), a declarative `mode` whose changes animate, and an imperative handle that reports when each transition lands. The choreography lives in a framework-agnostic engine (`processingEngine.ts`); the React component only renders markup and forwards props, so a web component could share it.

#### Changelog

**New**

- `mode` is the state to be in (`'loading' | 'triangle' | 'square' | 'out'`). On mount the dots load in and go there; later changes animate from wherever the dots are. `onTransitionEnd(mode)` reports each state reached.
- Shape formations: the dots arc into a triangle or a square (a fourth dot grows in from the centre) while the group spins up into a steady rotation.
- Imperative handle via `ref`: `triggerTriangle()`, `triggerSquare()`, `triggerLoading()`, `triggerWiggle()`, `triggerOut()`. Each returns `Promise<boolean>` (true once reached, false if superseded). Requests made mid-transition queue, and the latest wins.
- `triggerLoading()` unwinds a shape back into the loading line as the same clockwise gesture every time, without interrupting the pulse. Triangle ↔ square passes through the line; after `out`, any request loads the dots in again.
- `label` is visually hidden text inside a `role="status"` live region, so changes are announced. New `ai` (color treatment) and `className` props.
- Reduced motion is handled in JavaScript (the Web Animations API ignores the CSS media query): the component paints static end states. The same fallback applies where the Web Animations API is unavailable.
- Right-to-left mirrors the drawing. Dot colors are per-theme component tokens: `--clabs-processing-dot-color` and `--clabs-processing-dot-color-ai`.

**Changed**

- The animation engine is the Web Animations API instead of CSS keyframes. Idle cost is one infinite animation per dot plus one for the spin; no JS runs between transitions.
- `loop` now repeats until the state changes (v1 ran five cycles, then unloaded). `loop={false}` still runs one cycle, then unloads.
- Stories: `Loading`, `Triangle`, `Square`, `Wiggle`, `Mode prop` and `Interactive controls v2.0 proposal`. Docs cover the API, handle, accessibility, RTL, theming and migration.
- `telemetry.yml` regenerated for the new props and `mode` values.

**Removed**

- The `__linear` / `__linear--no-loop` modifier classes and their keyframes, and the `__dot--left` / `--center` / `--right` modifiers (dots now swap line positions as shapes unwind).
- The `LinearLoop` / `LinearNoLoop` stories (their story IDs change).
- Rendering without JavaScript: the dots are drawn by JS, so nothing is visible before hydration.

#### Testing / Reviewing

- `cd packages/react && yarn storybook`, then open **Components/Processing**:
  - **Interactive controls v2.0 proposal**: try Triangle, Square, Loading, Wiggle and Out, including pressing a second button mid-transition (it queues).
  - **Mode prop**: change `mode` in the controls and watch it animate; `onTransitionEnd` appears in the Actions panel.
- Turn on reduced motion in the OS: shapes appear as static end states, with no load-in, pulse or spin.
- Set `dir="rtl"` on a parent: the drawing is mirrored.
- `npx jest src/components/Processing` (in `packages/react`): 15 tests, including engine sequencing, queueing, the unwind's role reassignment and reduced motion, driven by a Web Animations API stand-in.
- `yarn ci-check` passes.


mikeolasov:

mikeolasov:
recheck

glapadre (changes requested):
Looks good overall. Small change, in Storybook when you go to a specific story and go to the code tab we should show the actual code being represented. For your new stories it's showing the storybook implementation instead. Should be a easy fix with Bob. 

For example it should show something like:

`<Processing ref={ref} label="Processing" />`

but instead getting a long Storybook block such as:

`{
  render: () => <Demo script={async (h, alive) => {
    await sleep(LOAD_IN_END + HOLD);
    if (alive()) {
      await h.triggerOut();
    }
    await sleep(400);
  }} />,
  parameters: {
    docs: {
      description: {
        story: 'The default loading state: three dots load in and pulse, then shrink out and restart.'
      }
    }
  }
}`

